PR #502, opened 2019-11-13
feat(subscribe-button): add a fake loader on subscribe button

lutangar:


newick:
We should delete this one, no need for it I think @lutangar @JalilArfaoui

lutangar:
I'm closing this, for button we shouldn't add a fake loading but a true.
If I remember correctly on the "profile" app this is handle by a true loader: 
As long as we don't receive the subscirbe success action, the button show a loading state"

MaartenLMEM:
@lutangar why didn't you just modify the title by deleting "fake" ?

PR #570, opened 2020-02-03
refactor(extension): set background script as non persistent and remove html template

lutangar:
background script do not use any persistent background feature, this was a legacy thing, this will definitely be a big take on performance improvement
https://developer.chrome.com/extensions/background_migration


JalilArfaoui:
@lutangar should we close this one ?

lutangar:
I think this would bring a significant performance boost (memory footprint) and might be easier on the cpu as well so I'd like to put this light. I'll might add a connected issue.

lutangar:
After some more investigation, switching to alarm might not be useful as in manifest v3 the background persistence thing seems to be deprecated... there might still be performant boost to take from switching to the alarms API tho.

This PR is still a WIP it'll require a bit more work to add some kind of alarm registry to make the onAlarm eventListener able to  handle all alarms...

PR #623, opened 2020-04-14
Feat/profiles

lutangar:


lutangar:
@JalilArfaoui this PR is supposed to be a base for @newick job on integration. 
I will not do the "external connection" to the extension here, and I'm creating another PR having this one as a target. This will be easier to review, work and why not, ship a first version without the extension interactions.
Anyway, this is ready for review, I still need to figure out the organization to separate the "profiles "app from the rest of the build.

src/app/store/collection/reducers/items.ts
```diff
@@ -0,0 +1,30 @@
...
+export const initialState = [];
+
+export default <I extends Item>(SUCCESS: string, itemIdentifier: string) => (
+  state: ItemsState<I> = initialState,
+  action: ItemsAction<I>
+): ItemsState<I> => {
+  switch (action.type) {
+    case SUCCESS: {
+      return R.uniqWith(
+        // @ts-ignore
```
JalilArfaoui, line 22: Solution was to ensure that 'Item' contains the 'itemIdentifier'. This works for me :

```ts
import * as R from 'ramda';

export type Item<Id extends string = 'id'> = Record<Id, unknown>;

export interface ItemsAction<I extends Item<Id>, Id extends string> {
  type: string;
  payload: I[];
}

export type ItemsState<I extends Item<Id>, Id extends string> = I[];

export const initialState = [];

export default <I extends Item<Id>, Id extends string>(
  SUCCESS: string,
  itemId: string
) => (
  state: ItemsState<I, Id> = initialState,
  action: ItemsAction<I, Id>
): ItemsState<I, Id> => {
  switch (action.type) {
    case SUCCESS: {
      return R.uniqWith(R.eqProps(itemId), R.concat(state, action.payload));
    }
    default:
      return state;
  }
};
```

JalilArfaoui: @lutangar I’ve applied my own solution but I’ve reverted it because it breaks other parts :-(

TypeScript is correct yelling at us :
- We define `ItemsState` as `<I extends Item> = I[];`, `Item` being a `{}` interface
- We then use this property in the reducer to be able to `R.eqProps(itemIdentifier),`
- But in `src/app/profiles/store/reducers/subscriptions.reducer.ts` we use `ItemsState<Subscription>`, `Subscription` being a `number`
- It only works because we don’t use the default `item` reducer from `collection` helper, but a custom one. 

But it’s still wrong as here `Subscription` **is not** a valid `Item` … and we may create a bug by confusing a `number` with an `object` with `id` …

I create an issue from this in order to merge it in current state

PR #631, opened 2020-04-24
fix(styling): Fix styled-component messing with host page

lutangar:
I was struggling with "Profiles page" styles, and I figure out we had similar issues before.
Even if we were targeting the iframe head to inject the styles, the previous version of `styled-component` was still injecting an empty <style> tag in the host page head. This tag was in fact messing *every* host page style which were using the same version of styled-component than ours. :scream:


JalilArfaoui:
> I was struggling with "Profiles page" styles, and I figure out we had similar issues before.
> Even if we were targeting the iframe head to inject the styles, the previous version of `styled-component` was still injecting an empty <style> tag in the host page head. This tag was in fact messing _every_ host page style which were using the same version of styled-component than ours. scream

😱 Thanks !

PR #635, opened 2020-05-08
chore: Use browser polyfill where relevant.

lutangar:


src/webext/createBrowserStorage.ts
```diff
@@ -2,50 +2,12 @@ import { WebStorage } from 'redux-persist/es/types';
...
-        storage.remove(key, () => {
-          if (chrome.runtime.lastError == null) {
-            resolve();
-          } else {
-            reject(new Error(chrome.runtime.lastError.message));
-          }
-        });
-      });
-    }
+    getItem: (key: string) => storage.get(key).then(value => value[key]),
```
JalilArfaoui, line 8: ♥👌

PR #641, opened 2020-06-04
chore: Reduce build size by importing core-js

lutangar:
Something seems wrong with our config, just with this change it seems the size of included core-js functions is reduced 4 times...

### DIFF
--- a/src/app/background/index.ts
+++ b/src/app/background/index.ts
@@ -1,5 +1,6 @@
 /* eslint-disable no-console */
 // Early imports with high priority stuff involved, such as event listeners creation
+import 'core-js/stable';
 import { BACKEND_ORIGIN } from 'app/constants/origins';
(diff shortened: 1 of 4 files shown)


lutangar:
I forgot to create this PR a while back. But I think we are just missing the `preset-env`  bit here, this would do the job for us with the right config.
> https://babeljs.io/docs/en/babel-preset-env#corejs

JalilArfaoui:
@lutangar I don’t see how it reduces the build size, since we’re still importing it another way … Did you test this with `yarn analyze`  ?

lutangar:
> Did you test this with yarn analyze

Yes I did, but things might have change since then. Maybe it would look less like webpack voodoo magic today....

JalilArfaoui:
@lutangar I’m still not convinced this makes any difference … Could you try to rebase and check ? Or do we close it ?

PR #645, opened 2020-06-24
fix(fonts): added missing fonts from google fonts

newick:


newick:
thanks for your feedbacks @lutangar, it works for me

PR #646, opened 2020-06-25
Fix/no contribution screen

newick:
This old card was not closed https://trello.com/c/2rYCPamv/468-simplifier-les-labels-quand-pas-de-contribution-sur-la-page

Btw, the code is was not on staging this day, does it disappear ?


JalilArfaoui:
@newick This is `not` on master.

I’ve just approved it !

JalilArfaoui:
Oh ! @lutangar was faster

PR #650, opened 2020-07-02
chore(extension-detection): Only post message if the location match t…

lutangar:
…he profiles origin


lutangar:
@JalilArfaoui could you review this little one please?

Issue #634, opened 2020-05-01
Issue with CGU validation


johnkrovitch:
Hi,

I have a strange bug with CGU validation : I need to validate the CGU multiple times. It is like I need to validate for each site managed by the extension.

For example, I browse amazon.fr, the extension asks me to validate the CGU. I click and do validate. Then I go to another site, airbnb.fr, then the extension asks again me to validate the CGU.

But the strange thing is that the bug does not appear every time. Sometime I validate the CGU once, and it works.

It is like the synchronization is not working everytime.

I'm using Firefox on Windows and Linux (KDE Neon), and use Firefox synchronization (history...) between each instance of Firefox.

Thanks

MaartenLMEM:
Thanks a lot John !
There is obviously a problem here. With a colleague, we tried to reproduce the bug, we didn't success.
But last days a dozen of people who uninstalled Dismoi told "it requested a new permission". 
Also one of us had same problem as you but just one time.
We are trying to understand and find the origin of it. 
If you agree we will contact you for some more details.
Best,

johnkrovitch:
@MaartenLMEM yes of course.

lutangar:
Thank you for reporting this issue! 
We'll need a few more informations before digging further :
- Approximate date of first installation
- Current Dismoi extension version number (check about:addons in Firefox)

I'll get back to you asap.

johnkrovitch:
Hi,
the extension version is 3.30.0 (and Firefox 76.0.1 (64 bits)).
The first installation was a long time ago, when it was called "le même en mieux". I do not know the precise date.

johnkrovitch:
Hello,

I may have new data on the bug : I had the issue in my Linux browser (Firefox 76.0.1). It seems that the extension tells me to validate the CGU only if there is a contribution on the current page. Otherwise, it seems that have do not have the issue.

lutangar:
After digging on lost settings it's seems to be a recurring issue with extensions in general. 

See the following :
- https://help.getadblock.com/support/solutions/articles/6000169060-i-keep-seeing-your-payment-page-or-adblock-keeps-losing-my-settings-in-firefox
- https://www.reddit.com/r/firefox/comments/bwv68f/how_to_stop_firefox_from_automatically_resetting/

We decided to change the design of the extension to fit this behavior which obviously we can't fully control.
We now consider that the ToS (CGU) have been accepted when the extension is installed.
> See more details in #666

We also made adjustments to make the extension painless to use without a state.

PR #666, opened 2020-07-20
feat: Consider ToS have been accepted when the extension is installed.

lutangar:
This will prevent the consent window appearing too many times for some users..

Background behind this are the browser storage persistence issues. Without an account it's difficult keep track of the user consent.
We saw other extensions having the same kind of trouble for persisting user's config. Overall, extensions tend to consider user installation as consent, but for us it was really an issue since we don't store personal data anyway.

This implementation is rather naive, it just consider that ToS are accepted and open example page on first install.
This is on purpose, and will allow us to reactivate ToS easily or use part of it elsewhere.

This is WIP to open the discussion.

> More context https://trello.com/c/EbHq7u0B/519-supprimer-tos

@JalilArfaoui @MaartenLMEM suggested to redirect to the profile page instead of the examples page and take the leap of faith :see_no_evil: 
It might be the right time to put profile app as the main options page.


lutangar:
@JalilArfaoui I updated the PR with the replacement of the options page for the profiles page.
I'm a bit a reluctant to remove all code related to options page. I'm also having second thought about not having a true options page, (can"t use an external page here, must be an internal page), but YEAH we have a context menu entry so it's not that bad.
But for example, since Firefox profile page is a bit wonky, we could for example run the profile page in the context of the options page, that should work out of the box. I might be overly careful about this...

src/app/background/index.ts
```diff
@@ -36,3 +37,18 @@ const handleConnect = (port: Port) => store.dispatch(connect(port));
 
 browser.runtime.onConnectExternal.addListener(handleConnect);
 browser.runtime.onConnect.addListener(handleConnect);
+
+// https://stackoverflow.com/questions/37000136/check-if-item-is-already-in-the-context-menu
+browser.contextMenus.remove(OPTIONS_MENU_ITEM_ID).then(() => {
```
JalilArfaoui, line 42: Starting the extension gives me an error on this line in the background : 

image

Maybe we have to either `removeAll` or maybe `try`/`catch` and `create` in the `catch` ?

lutangar:
Screenshot from 2020-09-04 17-15-04
@MaartenLMEM @JalilArfaoui Here's what it looks like. The weird thing to notice is the 2 "Options" menu entry, since we cannot use an external page for options, we need another "Options" entry.
An alternative, could be to also ship the "Profiles" app with extension, but that would defeat the added benefit of the clickable browser action. @MaartenLMEM wanted the browser action to be interactive on the options page.

lutangar:
Options app has been removed in favor of the profiles app.
And this card has been dealt with :
https://trello.com/c/r6W0G5pC/520-si-je-clique-sur-le-nom-du-contributeur-dans-lextension-sa-page-profil-souvre-dans-nouvel-onglet-plus-de-profil-dans-lextension

JalilArfaoui (approved):

src/app/background/selectors/prefs.ts
```diff
@@ -9,10 +9,7 @@ import { BackgroundState } from '../reducers';
 
 export const getPrefs = (state: BackgroundState) => state.prefs;
 
-export const areTosAccepted = createSelector(
-  getPrefs,
-  prefs => prefs.tosAccepted
-);
+export const areTosAccepted = createSelector([], () => true);
```
JalilArfaoui, line 12: Maybe we could remove this selector ?

src/app/content/App/Subscriptions/withConnect.ts
```diff
@@ -1,22 +1,22 @@
...
+import clickContributor from '../../actions/goToContributor';
 import { ContentState } from '../../store';
 import {
   getNbTotalContributors,
   getSubscriptions
-} from 'app/options/store/selectors/contributors.selectors';
-import { push } from 'connected-react-router';
+} from 'app/store/selectors/contributors.selectors';
+
+const openSubscriptions = () => optionsRequested();
```
JalilArfaoui, line 10: This can be simplified

src/components/organisms/Contributor/ContributorCompact.tsx
```diff
@@ -34,21 +34,30 @@ interface Props<S = LocationState> {
...
   onUnsubscribe: () => void;
   to?: LocationDescriptor<S>;
+  handleNameClick?: () => void;
   className?: string;
 }
 export const ContributorCompact = ({
   contributor,
   onSubscribe,
   onUnsubscribe,
+  handleNameClick,
```
JalilArfaoui, line 44: I usually find `onClick` to be more idiomatic in props, rather than `handleClick` which sounds more implementation related …

Here we could go with `onNameClick` for instance …

src/components/organisms/Contributor/ContributorCompact.tsx
```diff
@@ -34,21 +34,30 @@ interface Props<S = LocationState> {
...
   to,
   className
 }: Props) => (
   <Wrapper className={className}>
-    <Avatar contributor={contributor} size="small" to={to} />
+    <Avatar
+      contributor={contributor}
+      size="small"
+      to={to}
+      handleClick={handleNameClick}
```
JalilArfaoui, line 53: … and here `onClick`

src/components/organisms/Contributor/ContributorLarge.tsx
```diff
@@ -44,6 +44,7 @@ interface ContributorLargeProps<S = LocationState> {
   onUnsubscribe: () => void;
   className?: string;
   children?: React.ReactNode;
+  handleNameClick?: () => void;
```
JalilArfaoui, line 47: Same …

src/components/organisms/Notice/Notice.tsx
```diff
@@ -55,7 +61,7 @@ interface Props {
   dismiss: (id: number) => void;
   confirmDismiss: (id: number) => void;
   undismiss: (id: number) => void;
-  clickContributor: (id: number) => void;
+  clickContributor: (contributor: Contributor) => void;
```
JalilArfaoui, line 64: Here it would be `onContributorClick` to stay consistent

src/components/organisms/Notice/Notice.tsx
```diff
@@ -159,14 +165,14 @@ export default class Notice extends PureComponent<Props, CountDownState> {
           ) : (
             <>
               <InteractiveAvatar
-                onClick={this.onContributorClicked}
+                handleClick={this.onContributorClicked}
```
JalilArfaoui, line 168: Same

src/components/organisms/Contributor/ContributorCompact.tsx
lutangar: <3

src/components/organisms/Contributor/ContributorLarge.tsx
lutangar: :+1:

src/components/organisms/Notice/Notice.tsx
lutangar: :+1:

src/components/organisms/Notice/Notice.tsx
lutangar: :+1:

src/components/organisms/Contributor/ContributorCompact.tsx
lutangar: :+1:

src/app/content/App/Subscriptions/withConnect.ts
lutangar: Nop, or an event would be passed as an argument to the fn, and this would make the fn optionsRequested breaks.

JalilArfaoui: True ! My bad !

lutangar:
Also make #634 less painful.

PR #667, opened 2020-07-21
fix: Browser action click not toggling UI

lutangar:
Since we remove the `activeTab` permission we lost access to the tab URL in some part of the browser API.
We have a little hack to get the tab URL thus it may not be such an issue.
However it's not very future proof and from my point of view (maybe I missed something but) the `activeTab` tab permission seems relevant in our case.

Here, my implementation choice assumes a tab may not have an URL.


lutangar:
@MaartenLMEM @JalilArfaoui I'll merge this because it's urgent, but we should discuss this more thoroughly on next meeting.

PR #682, opened 2020-08-26
feat(profile): strip protocol from example URL

JalilArfaoui:


src/app/utils/stripUrlProtocol.ts
```diff
@@ -0,0 +1,2 @@
+export const stripUrlProtocol = (url: string): string =>
+  (u => u.host + u.pathname)(new URL(url));
```
lutangar, line 2: :+1:

lutangar (approved):

PR #685, opened 2020-08-31
feat(profiles): add pagination to contributor’s notices list

JalilArfaoui:


lutangar (changes requested):
I'm not a fan of the overall approach, here are a few feedback before discussing this further :

1. State :
```
type OffsetInfo = {
  current: number;
  fetchedAll: boolean;
};
```
The `fetchedAll` feels cumbersome, the required info should be the total number of items. The backend could provide this info on the first fetch.

2. Vocabulary
Instead of speaking of an `offset` I would rather use a vocabulary closer to the pagination :
`page` or `currentPage` looks more appropriate.

3. `fetchMoreContributorNotices` Couldn't we use the `fetchContributorNotices` with just an extra parameter ?

Also I wonder to what a pagination is tied to. Should an entity collection only handle a general pagination ? Or shall it be tied to a component only ? What if we need different pagination, filters, etc... in different part of the application ?

lutangar (approved):

PR #687, opened 2020-09-04
fix(banner): Fix default banner not showing up.

lutangar:
Default image rule and the profile one were someway conflicting, explicitly include profiles app directory does fix an issue with the image being corrupted.


lutangar:
@JalilArfaoui alright, I fixed the commit, this is now reviewable.

JalilArfaoui (approved):

PR #690, opened 2020-09-07
fix(firefox): When navigating to another page wrong tab URl was transmited

lutangar:
https://trello.com/c/I8tHGr5P/586-bulle-qui-persiste-sur-youtube-apres-avoir-mis-du-temps-%C3%A0-apparaitre

### DIFF
--- a/src/webext/createMessageHandler.ts
+++ b/src/webext/createMessageHandler.ts
@@ -1,6 +1,6 @@
 import { Action } from 'redux';
 import * as R from 'ramda';
-import { BaseAction } from 'app/actions';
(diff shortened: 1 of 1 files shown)


JalilArfaoui:
I understand that the `NAVIGATED_TO_URL` was sent from `content` before the URL had actually changed, so that `sender.url` was incorrect. Am I right ?

lutangar:
Could be, it was 100% reproductible on Youtube with Firefox, but not on Chrome, so might be a browser interpretation issue.

JalilArfaoui (approved):

PR #691, opened 2020-09-08
fix(subscribe): Fix subscription to a similar contributor.

lutangar:


src/app/profiles/App/Profiles/Profile/withConnect.ts
```diff
@@ -54,17 +54,10 @@ const mergeProps = (
 ) => ({
   ...ownProps,
   ...stateProps,
+  ...dispatchProps,
   fetchedAll: stateProps.contributor
     ? stateProps.fetchedAll(stateProps.contributor?.id)
     : false,
-  subscribe: () => {
```
JalilArfaoui, line 60: Can you tell me what was not working here ? Or is it an approach matter ?

lutangar: I think you previously changed this somewhere in the pagination, but I'm not sure you needed to, correct me if I'm wrong.
But the thing is it introduce a bug, since the subscribe method is also used on the "Similar profile component", all subscriptions done via this component were subscribing to the main informer instead ^ ^

JalilArfaoui: Yes, I get it, thanks

JalilArfaoui (approved):

PR #692, opened 2020-09-10
Fix/profiles/seo

newick:
replaces headings values for profiles


src/components/atoms/UserName/UserName-compact.ts
```diff
@@ -0,0 +1,25 @@
+import styled from 'styled-components';
```
JalilArfaoui, line 1: Why not `CompactUserName.ts` ?

newick: Good one, the habit of CSS…

lutangar (approved):

PR #693, opened 2020-09-14
fix: Setup profiles app proding and staging envs

lutangar:


webpack.profiles.config.js
```diff
@@ -89,16 +89,18 @@ module.exports = function webpack(env = {}, argv = {}) {
...
-            : {
-                from: 'test/integration',
-                to: path.join(
-                  defaultWebpackConfig.output.path,
-                  'test',
-                  'integration'
-                )
-              }
+          {
+            from: 'test/integration',
```
JalilArfaoui, line 93: Maybe we can remove the `test/integration` entry ? 

When is the last time it has been used ?

JalilArfaoui (changes requested):

webpack.profiles.config.js
lutangar: I still need `test/integration/bulle-extension.js` to deploy a fully functionnal environment when it's not published on the Wordpress.

PR #698, opened 2020-09-24
Feat/install popin

newick:
@lutangar you can plug it with data


lutangar:
@newick 
I rework this PR to reuse previously created components: InteractiveAvatar and list of avatar used on the "subscriptions" screen of the extension. I changed the margin used there to the margin used on the new list of avatar in the popin. Looks fine from point of view but I'll need another look from you before before merging this.

newick:
@lutangar for me the design is ok. 🚀

lutangar (approved):

Issue #272, opened 2019-05-10
[Storybook] Use knobs instead of hard-coded values


JalilArfaoui:

christpet:
Not sure what this means... unless you have an objection @JalilArfaoui I'll close this one for now.

PR #700, opened 2020-09-25
feat(contributor): Allow contributors to be filtered by categories

lutangar:
> Rely on backend PR [#272](https://github.com/dis-moi/backend/pull/272)

### DIFF
--- a/src/app/actions/categories.ts
+++ b/src/app/actions/categories.ts
@@ -1,0 +1,35 @@
+import { createErrorAction } from './helpers';
+import { Categories } from 'app/lmem/category';
+
+export const FETCH_CATEGORIES_REQUEST = 'CATEGORIES/FETCH_REQUEST';
+export const fetchCategoriesRequest = () => ({
+  type: FETCH_CATEGORIES_REQUEST
+});
+export type FetchCategoriesRequestAction = ReturnType<
+  typeof fetchCategoriesRequest
+>;
+
+export const FETCH_CATEGORIES_SUCCESS = 'CATEGORIES/FETCH_SUCCESS';
+export const fetchCategoriesSuccess = (categories: Categories) => ({
+  type: FETCH_CATEGORIES_SUCCESS,
+  payload: categories
+});
+export type FetchCategoriesSuccessAction = ReturnType<
+  typeof fetchCategoriesSuccess
+>;
+
+export const FETCH_CATEGORIES_FAILURE = 'CATEGORIES/FETCH_FAILURE';
+export const fetchCategoriesFailure = createErrorAction(
+  FETCH_CATEGORIES_FAILURE
+);
+export type FetchCategoriesFailureAction = ReturnType<
+  typeof fetchCategoriesFailure
+>;
+
+const types: [string, string, string] = [
+  FETCH_CATEGORIES_REQUEST,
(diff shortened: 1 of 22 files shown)


src/app/profiles/store/reducers/categories.reducer.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Categories } from 'app/lmem/category';
+import actionTypes, {
+  FETCH_CATEGORIES_SUCCESS,
+  FetchCategoriesSuccessAction
+} from 'app/actions/categories';
+
+export interface CategoriesCollection {
+  fetched: FetchedState;
+  fetching: FetchingState;
+  items: Categories;
```
JalilArfaoui, line 19: Why do we use an indexed collection, instead of a simple list, and thus we can’t use the generic`items` ?

src/app/profiles/App/useContributorsFilters.ts
```diff
@@ -0,0 +1,43 @@
+import { useState, useEffect } from 'react';
+import { StatefulContributor } from 'app/lmem/contributor';
+import { Categories } from 'app/lmem/category';
+
+const distinct = (value: string, index: number, array: string[]) =>
+  array.indexOf(value) === index;
+
+function useContributorsFilters(
```
JalilArfaoui, line 8: ❤️

src/app/profiles/store/reducers/categories.reducer.ts
lutangar: Because at the moment the list of categories is an enum on the backend side, so here it's a map and not a list, see :
https://staging-notices.bulles.fr/api/v3/categories

I'm open to discuss this format if you think it's not ideal.

src/app/actions/categories.ts
```diff
@@ -0,0 +1,40 @@
+import { BaseAction, ErrorAction } from './index';
+import { createErrorAction } from './helpers';
+import { Categories } from 'app/lmem/category';
+
+export const FETCH_CATEGORIES_REQUEST = 'CATEGORIES/FETCH_REQUEST';
```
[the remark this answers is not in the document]
lutangar: done

src/app/profiles/store/reducers/categories.reducer.ts
lutangar: ping @JalilArfaoui

JalilArfaoui: I meant that we could store them in Redux side as a list of `{ id: string; label: string; }` for instance, and thus use the generic `items` reducer …

No big deal here …

lutangar: > we could store them in Redux side as a list of { id: string; label: string; }
Agreed, this the way we store other stuff, and we might move to this structure once the category is more than a map.
But in fact I do like the fact that this is a map here.

Anyway, does it mean you are approving this?

lutangar: Sorry @JalilArfaoui but since this is a blocker for other features, I'm merging this.

JalilArfaoui: Yep, sorry, I did not see you responses … I’m fine with this !

PR #710, opened 2020-10-13
chore: Change subscriptions page URL to /abonnements

lutangar:
Don't forget to add the proper redirections @MaartenLMEM


MaartenLMEM:
This was my idea, but actually mes-abonnements was more explicite url for users and there won't be to children url behind

PR #714, opened 2020-10-19
Update FiltersSelect.tsx

MaartenLMEM:


lutangar:
@MaartenLMEM this component is actually not used yet. It will be used on the mobile version.
If you want to change the categories it must be done on the backend side :
https://github.com/dis-moi/backend/blob/master/src/Domain/Model/Enum/CategoryName.php

MaartenLMEM:
Ouch, I see an issue to manage "Tous". Probably easy for you ^^

lutangar:
Closing in favor of #715

PR #724, opened 2020-10-29
fix: Fix two scrollbars appearing on notice details since preview scr…

lutangar:
…ollbar fix


lutangar:
@newick I removed the margin-top, seems legit to me since the preview is supposed to look liked  a notice details... or might need to add it again everywhere, but then the preview height need to be changed for the button to be inside the container.

JalilArfaoui (approved):

PR #727, opened 2020-11-06
feat: Add campaign param when redirecting to profiles for the first time

lutangar:
I removed duplicated code for building a query string (for the ramda part I'm open to suggestion.
But I could as well keep the library instead of the custom function since it was there anyway and migth handl edge cases?
Or there's also the native URL and params browser API...

### DIFF
--- a/src/api/call.ts
+++ b/src/api/call.ts
@@ -4,34 +4,39 @@ import * as R from 'ramda';
 import { APIStatusCodeError } from './APIStatusCodeError';
 
 type GetParamValue = string | string[];
-type GetParams = {} | { [key: string]: GetParamValue };
+export type GetParams = {} | { [key: string]: GetParamValue };
 
 type BuildQueryString = (params: GetParams) => string;
-const buildQueryString: BuildQueryString = R.pipe(
-  R.toPairs,
-  R.map(([key, value]: R.KeyValuePair<string, GetParamValue>) =>
-    Array.isArray(value)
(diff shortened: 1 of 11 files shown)


src/api/call.ts
```diff
@@ -23,15 +23,19 @@ const buildQueryString: BuildQueryString = R.pipe(
   R.concat('?')
 );
 
+export const buildQueryString = (data: GetParams) =>
+  R.isEmpty(data) ? '' : buildDumbQueryString(data);
+
```
JalilArfaoui, line 28: If we want to avoid `buildDumbQueryString` we can write :

```ts
type BuildQueryString = (params: GetParams) => string;
const buildQueryString: BuildQueryString = R.ifElse(
  R.isEmpty,
  R.always(''),
  R.pipe(
    R.toPairs,
    R.map(([key, value]: R.KeyValuePair<string, GetParamValue>) =>
      Array.isArray(value)
        ? value
            .map(
              (valueItem: string) =>
                `${encodeURIComponent(key)}[]=${encodeURIComponent(valueItem)}`
            )
            .join('&')
        : `${encodeURIComponent(key)}=${encodeURIComponent(value)}`
    ),
    R.join('&'),
    R.concat('?')
  )
);
```

lutangar: :+1:

JalilArfaoui (approved):

PR #728, opened 2020-11-06
fix: Fix crash when trying to access "my subscriptions" page while no…

lutangar:
…t connected
Screenshot from 2020-11-06 14-40-03


JalilArfaoui:
What was the issue ?

lutangar:
@jalil 
```
react-dom.development.js:55 Uncaught Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.
```

JalilArfaoui (approved):

Issue #362, opened 2019-09-06
TS hook doesn't produce same error as TS in watch mode


lutangar:
```
src/app/background/reducers/index.ts:8:32 - error TS7016: Could not find a declaration file for module 'redux-persist/es/types'. '/home/lutangar/workspace/recommendations-webextension/node_modules/redux-persist/es/types.js' implicitly has an 'any' type.
  Try `npm install @types/redux-persist` if it exists or add a new declaration (.d.ts) file containing `declare module 'redux-persist/es/types';`

8 import { PersistedState } from 'redux-persist/es/types';
                                 ~~~~~~~~~~~~~~~~~~~~~~~~
```
When attempting to push I got the following error from typescript, but not running `yarn start`.
There is a related issue https://github.com/rt2zz/redux-persist/issues/700 BUT in our project everything seems to run fine everywhere expept when attempting to push.

Might be configuration diff between the ways TS is ran.

lutangar:
@JalilArfaoui previous comment is a bit outdated, but at the time it was probably different output because we used either `tsc --noEmit` or `eslint 'src/**/*.{ts,tsx}'`.
Still questions persist :
- Is `tsc --noEmit` still relevant?
- If so, shouldn't we include this typecheck in the lint command or do you consider this as a different job?

JalilArfaoui:
@lutangar to me : 

- `tsc --noEmit` checks the correctness of typings and their usage. It does not case about the code style
- `eslint 'src/**/*.{ts,tsx}'` checks the code style of all JavaScript *and* TypeScript files, including the typings style

They are both useful to me on their respective job.

Typing may be detected as incorrect (such as in the description), and the application still building correctly, as we use `fork-ts-checker-webpack-plugin` to have concurrent non-blocking type checking …

But you should still see the type error in the build console, right ?

JalilArfaoui:
@lutangar Do you still have an issue with the setup ? 
How can I reproduce ?

lutangar:
Regarding #729  
I just stumble upon another reason for the strange feeling: 
**Typechecking isn't run on the test folder when watching for type errors*

Example:
1. I commit smthing.
2. Run `yarn typecheck` and saw an error in `test/fakers/generateContributor.ts`

JalilArfaoui:
Do we want to always check types in /test (watch+verify) of never ? (true question)

lutangar:
I'd say always. What about you?

JalilArfaoui:
[...]

PR #729, opened 2020-11-08
chore: verify(all) on pre-push hook

JalilArfaoui:
This should help with #362 : 

Instead of just running tests on pre-push, we now check : 

- tests
- typings
- code and style lint 
- build of extension and storybook

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,28 +42,28 @@
     "release": "yarn semantic-release",
     "build": "yarn build:staging && yarn build:production && yarn build:proding",
     "clean": "rm -rf build",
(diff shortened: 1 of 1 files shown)


JalilArfaoui:
@lutangar if you think this PR already makes things better, we can merge it and test drive it in real life …

JalilArfaoui:
@lutangar ?

lutangar:
I'm reading #362 again...
There are still some fishy stuff on the test files ex: 
- linting does not seems to include test files
- same things for for the lint staged on pre-commit

I'm still not unsure why aren't running some more checks on pre-commit but let's talk about that another time :D

PR #730, opened 2020-11-09
Feat/more pinned contributions

lutangar:
Screenshot from 2020-11-09 15-38-21


lutangar:
@JalilArfaoui while I am at it, I'm a bit puzzled by all the variety of keyword used :
- `example` in the past 
- then `starred` but `featured` in the extension
- now `pinned`

If I had to choose I'll go for **featured** everywhere : "Contributions phares"
> https://www.deepl.com/fr/translator#en/fr/featured

Another topic but we might as well remove entirely the `contribution` "namespace" from the contributor model in the future.

JalilArfaoui:
> I'm a bit puzzled by all the variety of keyword used :
> 
>     * `example` in the past
> 
>     * then `starred` but `featured` in the extension
> 
>     * now `pinned`
> 
> 
> If I had to choose I'll go for **featured** everywhere : "Contributions phares"

Let’s discuss the (one-and-only) choice tomorrow with @MaartenLMEM, but we definitely have 2 much words. Historically, it made sense : `example` was used when automatically defined, the others are more recent.

To me : 
- `featured` implies an editor’s choice among users publications --> True today
- `pinned` (and previously `starred`) implies as selection from each contributor --> True tomorrow

>  we might as well remove entirely the `contribution` "namespace" from the contributor model in the future.

Totally ! I never got the meaning of this …

JalilArfaoui (approved):

PR #732, opened 2020-11-12
Excluded patterns and restricted contexts

lutangar:
While exploring the browser action title feature, I had to go through all the *exclusion* mechanism again and I produce a bit of doc in the process.

@MaartenLMEM please read this carefully.

This is also interesting, from the extension perspective, for the [workflow issue in the backend](https://github.com/dis-moi/backend/pull/287). 
> By the way I think we should have a similar schema in the extension repository.

### DIFF
--- a/docs/EXCLUSION.md
+++ b/docs/EXCLUSION.md
@@ -1,0 +1,93 @@
+# Matching exclusions
+
+There are different level of pages/contents excluded from the matching engine.
+
+> Most of these mechanisms use a *blacklist* for the time being. 
+> If possible we'd love to put in place a *whitelist* system (see [#275](https://github.com/dis-moi/extension/issues/275))
+
+## Browser
+For example, `chrom*` browsers already forbid all content scripting on the following pages:
+- new tab page
+- `chrome://` URLs
+- chrome web-store URLS
+
+## Manifest file
+There is a list of pattern exclusion in the extension of the manifest: 
+It doesn't inject the content script in pages matching those URLs patterns.
+Thus, it doesn't run the matching engine against these URLs.
+
+At the time of writing this consists in the following patterns:
+```
+exclude_globs: [
+  '*.pdf*',
+  '*.Pdf*',
+  '*.PDF*',
+  '*.jpeg*',
+  '*.jpg*',
+  '*.png*',
+  '*.gif*'
+],
+exclude_matches: [
+  '*://*.googleusercontent.com/viewer/secure/pdf/*',
+  '*://*.cdn.mozilla.net/*',
+  '*://*.accounts.firefox.com/*',
+  '*://*.addons.mozilla.org/*',
+  '*://*.input.mozilla.org/*',
+  '*://*.install.mozilla.org/*',
+  '*://*.support.mozilla.org/*',
+  '*://*.services.mozilla.org/*',
+  '*://*.testpilot.firefox.com/*'
+],
+```
+> You can find an up to date list in the  [`manifest/base.js`](../src/webext/forbiddenTabs.ts) file.
+
+Some of these rules cames from the [list provided by Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts) 
+and were also added to the `chrom*` version of the extension.
+Thus, URLs excluded from Firefox are also excluded from `chrom*` to provide a consistent user experience across browsers.
+
+## Coded exclusion list
+Here the extension injects the content script in the page __but__ it doesn't run the matching engine.
+Thus, we don't collect any data on these URLs.
+
+### Hard coded exclusion list
+We maintain a list of excluded patterns in the code.
+It is maintained irregularly and, some of these patterns might end up in the manifest file.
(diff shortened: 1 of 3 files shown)


docs/EXCLUSION.md
```diff
@@ -0,0 +1,93 @@
+# Matching exclusions
+
+There are different level of pages/contents excluded from the matching engine.
+
+> Most of these mechanisms use a *blacklist* for the time being. 
+> We think about putting in place a *whitelist* system (see [#275](https://github.com/dis-moi/extension/issues/275))
```
JalilArfaoui, line 6: We’re thinking about

docs/EXCLUSION.md
```diff
@@ -0,0 +1,93 @@
...
+  '*://*.accounts.firefox.com/*',
+  '*://*.addons.mozilla.org/*',
+  '*://*.input.mozilla.org/*',
+  '*://*.install.mozilla.org/*',
+  '*://*.support.mozilla.org/*',
+  '*://*.services.mozilla.org/*',
+  '*://*.testpilot.firefox.com/*'
+],
+```
+> You can find an up to date list in the `manifest/base.js` file.
```
JalilArfaoui, line 42: Maybe insert a direct link to the file on master ?

docs/EXCLUSION.md
```diff
@@ -0,0 +1,93 @@
...
+
+Some of these rules cames from the [list provided by Firefox](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_scripts) 
+and are also added on the `chrom*` version of the extension,
+to provide a consistent user experience across browsers.
+
+## Coded exclusion list
```
JalilArfaoui, line 48: Unclear … What are you thinking of ?

src/app/background/sagas/watchActivatedTab.saga.ts
```diff
@@ -27,7 +27,7 @@ export default function*() {
 
       if (!tab || !tab.url) {
         yield call(resetBadge, tabId);
-        yield call(disable, tabId);
+        yield call(disable, tab);
```
JalilArfaoui, line 30: Is that change voluntary ?

lutangar: Completely, see 1bc41efbb74f45d1425401b29f80352ed1209525

docs/EXCLUSION.md
lutangar: I rephrased

docs/EXCLUSION.md
lutangar: rephrased

MaartenLMEM:
Thanks for the overview @lutangar 
Our priority targets are "Restricted contexts" and "Hard coded exclusion list". 
Let's talk about it next days.

docs/EXCLUSION.md
```diff
@@ -0,0 +1,93 @@
...
+and were also added to the `chrom*` version of the extension.
+Thus, URLs excluded from Firefox are also excluded from `chrom*` to provide a consistent user experience across browsers.
+
+## Coded exclusion list
+Here the extension injects the content script in the page __but__ it doesn't run the matching engine.
+Thus, we don't collect any data on these URLs.
+
+### Hard coded exclusion list
+We maintain a list of excluded patterns in the code.
+It is maintained irregularly and, some of these patterns might end up in the manifest file.
```
MaartenLMEM, line 54: It might be related to pdf issues we had in the past.

MaartenLMEM:
Following our discussion @lutangar 

## Tooltip : just a quick win not solution
- Tooltip is not a remedy, since on Chrome there is always mentionned "A accès à ce site".
Nevertheless, there a quich win possible here : add "DisMoi restreint" on toolitip instead of just DisMoi on each Restricted context" and page excluded by Hard coded exclusions.

##Options 
###1/ Display a custom message in content script on each restricted page when user clicks on browser action (today nothing happens if he/she clicks in this restrictred context).
_The limit of this option is that is won't be displayed in a part of restricted context : Manifest file exclusions, especially new window (empty) situation (which is very frequent after having installed the add-on)_

###2/ Use the native pop-up of Chrome and Firefox API to add a message
Its likely to the Honey add-on strategy : 
A/ Native pop-up displayed when user clicks on browser action
B/ Content script pop-up diplayed when context is matched by a notice 

In this pattern, discrimantory factor on DisMoi can be : 
- option I : a notice exist for the current page. If yes, content script pop-up no (automaticcaly or after user click on browser action), if no then native pop-up (user click on browser action)
- option II : Native pop is dedicated to all restricted context. 
Option II seems me the best one, because if we are in a retricted contect, the basic message to user is "Here DisMoi is idle, there is no message and there won't be"

christpet:
@lutangar can this be merged or should we toss it out?

JalilArfaoui (commented):
@lutangar looks great to me, except I don’t get the difference you make between `Coded exclusion list` and `Hard coded exclusion list`

lutangar:
> Coded exclusion list

I removed this mention, I don't know what it was about maybe a confusion with the restricted contexts.

JalilArfaoui (approved):

PR #733, opened 2020-11-12
chore: separate SENTRY_ENABLED and SENTRY_SEND_VERSION

JalilArfaoui:
We want to be able to generate a build with Sentry integrated and enabled without uploading the version to the Sentry platform.

`SENTRY_ORG`, `SENTRY_PROJECT`, `SENTRY_DSN` and `SENTRY_ENABLED` are now committed in staging and production .env file because they are no secret, and we know the values we want to use depending on the environment.

New var `SENTRY_SEND_VERSION` is not defined in any .env file but forced by npm script when using `buildVersion` instead of `build` script, thus triggering the upload of the built version to Sentry service. Such operation does need a .sentryclirc file with the user `token` to complete.


lutangar (changes requested):
<3 it.... but could you add your description to the doc? :D

JalilArfaoui:
> <3 it.... but could you add your description to the doc? :D

Ha ha ! I was thinking you would say that ! Is it more `CONTRIBUTING` or `README` stuff to you ?

lutangar (approved):

lutangar:
If we speak about env somewhere, drop it there. Otherwise drop a new file in the docs folder.

lutangar:
Or a file dedicated to store publishing!

lutangar (approved):
<3

PR #734, opened 2020-11-13
fix(profile): Fix avatar not looking square

lutangar:


src/app/profiles/App/Pages/Profiles/Profile/ProfileIntro.tsx
```diff
@@ -43,8 +43,8 @@ const ProfileIntroContent = styled.section`
 
         ${Avatar} {
           margin-top: 0;
-          width: 70px;
```
JalilArfaoui, line 46: Do you know why this was malforming images ?

lutangar: Not in details but border radius 50% and size restriction of the container were responsible.

newick:
It comes from WordPress which has a 
```
img {
max-width: 100%;
}
```

if we reset `max-width: initial`, it works

JalilArfaoui (approved):

lutangar:
Updated with @newick solution!

JalilArfaoui (approved):
💕

PR #736, opened 2020-11-18
feat(matcher): Allow to match a page based on its content

lutangar:
> https://github.com/dis-moi/backend/pull/295

### DIFF
--- a/src/app/lmem/matchingContext.ts
+++ b/src/app/lmem/matchingContext.ts
@@ -1,4 +1,6 @@
 import { captureException } from '../utils/sentry';
+import Tab from './tab';
+import { sendContentScriptRequest } from 'webext/contentScript';
 
 export interface MatchingContext {
   id: number;
@@ -7,6 +9,7 @@ export interface MatchingContext {
   urlRegex: string;
   excludeUrlRegex?: string;
   querySelector?: string;
+  xpath?: string;
 }
 
 export interface RestrictedContext {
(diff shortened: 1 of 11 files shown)


src/app/lmem/matchingContext.ts
```diff
@@ -33,10 +36,30 @@ export const urlMatchesContext = (
   }
 };
 
-export const findMatchingOffersAccordingToPreferences = (
+export const filterContextsMatchingUrl = (
```
JalilArfaoui, line 39: @MaartenLMEM should we allow a specific notice to match only on a xpath with no domain/url specified ?  (Example : Any page containing «Maarten Samson» in a heading) ?

If yes, I suggest to change `urlMatchesContext` to :

```ts
export const urlMatchesContext = (
  url: string,
  { urlRegex, excludeUrlRegex }: MatchingContext
): boolean => {
  try {
    return (
      (!urlRegex || new RegExp(urlRegex, 'i').test(url)) && // <- here we allow bypassing `urlRegex`
      !(excludeUrlRegex && new RegExp(excludeUrlRegex, 'i').test(url))
    );
  } catch (err) {
    if (process.env.NODE_ENV !== 'test') {
      captureException(err);
    }
    return false;
  }
};
```

Warning : such notices may have an energy/performance impact thought, because xpath would be tested on every page seen …

JalilArfaoui (approved):
Very fan of what you did here ❤

Did you think of a way to re-test the xpath when the content of the page change without the URL changing ? (Typical SPA …)

src/app/lmem/matchingContext.ts
MaartenLMEM: > @MaartenLMEM should we allow a specific notice to match only on a xpath with no domain/url specified ?  (Example : Any page containing «Maarten Samson» in a heading) ?
> 
> Warning : such notices may have an energy/performance impact thought, because xpath would be tested on every page seen …

Yes I think we should we allow a specific notice to match only on a xpath with no domain specified (or group of domains of course)
Because there are some cases we need to target websites that we don't know, for example fake sites that imitate a big brand site to get user Bank Card number et that uses some patterns that are recognizables in the page  content and impossible to predict url that will be used by this m... f....

JalilArfaoui: To be more precise, we’re talking about matching the page content (xpath) without limiting neither to a domain **nor an URL pattern**

MaartenLMEM: You are right @JalilArfaoui . I didn't mention _URL pattern_ in my last answer sentence, but yes it s the same conclusion : contributor can specify a page content part without domain nor an URL pattern.

MaartenLMEM: OK juste had a talk with @JalilArfaoui and I ^had to admit there is no current usecase with url matching (it s just maybe/probably in the future)
So @lutangar current state of code is ok

lutangar: Good to know :+1: 
I agree this feature could be useful, but again this might go too far. The URLs pattern thing

PR #739, opened 2020-11-20
fix(featured): Introduce partial notice data model

lutangar:
> Depends on https://github.com/dis-moi/backend/pull/296

Fix  https://trello.com/c/IOThAQOA/565-jusqu%C3%A0-5-contributions-phares-%C3%A0-un-m%C3%AAme-profil

### DIFF
--- a/src/app/lmem/contributor.ts
+++ b/src/app/lmem/contributor.ts
@@ -1,4 +1,5 @@
 import * as R from 'ramda';
+import { NoticeItem } from './notice';
 
 export type AvatarSize = 'small' | 'normal' | 'large';
 export type Avatar = {
@@ -11,11 +12,8 @@ interface Picture {
 
 export type ContributorId = number;
 
-export interface PinnedNotice {
+export interface PinnedNotice extends NoticeItem {
   sort?: number;
-  exampleMatchingUrl: string;
-  noticeId: number;
-  noticeUrl: string;
 }
 
 interface Contribution {
--- a/src/app/lmem/contributor.ts
+++ b/src/app/lmem/contributor.ts
@@ -17,8 +17,7 @@ export interface PinnedNotice extends NoticeItem {
 }
 
 interface Contribution {
-  example: PinnedNotice; // @deprecated use `pinnedNotices` instead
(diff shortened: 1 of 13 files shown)


src/app/lmem/notice.ts
```diff
@@ -3,6 +3,16 @@ import { Ratings } from './rating';
 import { Contributor, ContributorId, NewContributor } from './contributor';
 import { captureMessage } from '../utils/sentry';
 
+export interface NoticeItem {
```
JalilArfaoui, line 6: Shouldn’t we try to merge `NoticeItem` and `BaseNotice` to one unique model ?

lutangar: Definitely! 
But can't do that until Friday though.

lutangar: Could have a second look at this, I had to update the API as well to get something neat:
> https://github.com/dis-moi/backend/pull/298

src/app/lmem/contributor.ts
```diff
@@ -11,16 +12,12 @@ interface Picture {
...
   sort?: number;
-  exampleMatchingUrl: string;
-  noticeId: number;
-  noticeUrl: string;
 }
 
 interface Contribution {
-  example: PinnedNotice; // @deprecated use `pinnedNotices` instead
-  starred: PinnedNotice; // @deprecated use `pinnedNotices` instead
+  example: NoticeItem;
```
JalilArfaoui, line 20: Do we still use this property ? If not, we could remove it since it’s deprecated …

lutangar: `example` it's not actually deprecated, I thought it was but it's not, but will certainly be in the future.

JalilArfaoui: Where is it used ?

lutangar: Don't remember at the moment, but I'll dig this and create an issue.
Could we merge this fix and discuss this later?

JalilArfaoui: Definitely, no blocker !

JalilArfaoui (approved):

PR #747, opened 2020-11-26
fix(DetailsContent): changes margins to improve readibility

### LINKED ISSUE #742: Margin missing on detailed message when there is the scrollbar

newick:
fixes #742


JalilArfaoui (approved):
LGTM

lutangar:
@newick I think the idea was to put a margin only when there is a scrollbar. Would that be possible?

MaartenLMEM:
My bad, Erick showed me th UI without scrollbar and I didn't give this feedback. 
It's a bit quibbling but to me also it seems better to keep text so large as possible when there is no scroll bar, in this case margin can be thiner.

lutangar (approved):

PR #758, opened 2020-12-09
Feature/respect do not track preference

felix-lambert:
I completely delete all the tracking requests link to `MatomoTracker`. So in production, if the user checks "Do not track" in the navigator settings, no tracking requests will be executed

### DIFF
--- a/src/app/background/sagas/index.ts
+++ b/src/app/background/sagas/index.ts
@@ -22,9 +22,12 @@ import { fetchRestrictedContextsSaga } from './fetchRestrictedContexts.saga';
 import connectSaga from './connect.saga';
 import installationDetailsSaga from './installationDetails.saga';
 import { loginSaga } from './user.saga';
+import { doNotTrackSettingFromNavigatorIsActivated } from '../../../webext/checkDoNotTrack';
 
 const tracker =
-  process.env.TRACKING_SITE_ID && process.env.TRACKING_URL
+  process.env.TRACKING_SITE_ID &&
+  process.env.TRACKING_URL &&
+  doNotTrackSettingFromNavigatorIsActivated() === false
     ? new MatomoTracker(process.env.TRACKING_SITE_ID, process.env.TRACKING_URL)
     : undefined;
 
--- a/src/webext/checkDoNotTrack.ts
+++ b/src/webext/checkDoNotTrack.ts
@@ -1,0 +1,9 @@
+export function doNotTrackSettingFromNavigatorIsActivated(): boolean {
+  const doNotTrack = window.doNotTrack || navigator.doNotTrack;
+
+  if (doNotTrack === '1' || doNotTrack === 'yes') {
+    return true;
+  }
+
+  return false;
+}


src/app/background/sagas/index.ts
```diff
@@ -24,7 +24,9 @@ import installationDetailsSaga from './installationDetails.saga';
 import { loginSaga } from './user.saga';
 
 const tracker =
-  process.env.TRACKING_SITE_ID && process.env.TRACKING_URL
+  process.env.TRACKING_SITE_ID &&
+  process.env.TRACKING_URL &&
+  navigator.doNotTrack === null
```
achauve, line 29: Great to improve data privacy :) But I think it would be better to use `navigator.doNotTrack !== 1` because the user could have opted in for tracking.
Looking at the spec, it might be either 1 (no tracking), 0 (tracking accepted), null or undefined.

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack

achauve (changes requested):

src/app/background/sagas/index.ts
felix-lambert: Completely true! Hope I will have the opportunity to work with you ;)

felix-lambert: Typescript was complaining about `navigator.doNotTrack !== 1`. By reading the doc more carefully I just realized that I did'nt apply well the condition well for all navigators and all versions. The new commit should work. Maybe we could think of having an end-to-end test system to be sure that all changes can apply to all navigators or not.

achauve: Indeed you're right, it's not `1` but rather `"1"` no?

Your code is really interesting and seems thorough, but as Dis-moi is only targetting Chrome and Firefox (am I wrong here?), I supppose adding a lot a code for IE is not worth the added complexity. I would rather try to keep things as simple as possible.

felix-lambert: Agreed

MaartenLMEM: I confirm the need : IE is not targeted. Just Firefox, Chrome and browsers based on those (Opera, Edge). Althought the latter  have a few users, so we don"t need to verify if it works with them.

JalilArfaoui (approved):
Looks good to me

Not directly related, but I would move the `tracker` definition inside the `trackingSaga` file (but outside of saga itself) to unclutter this index file …

src/webext/checkDoNotTrack.ts
```diff
@@ -0,0 +1,9 @@
+export const doNotTrackSettingFromNavigatorIsActivated = () => {
```
achauve, line 1: Some typescript details :
- add return type hint as the function is exported (seems even a good practice to me to always annotate function types)
- why not use a function instead of a variable assignation for a function declaration?
- I would simplify with `return doNotTrack === '1' || doNotTrack === 'yes'`, but it's really really minor

felix-lambert: > add return type hint as the function is exported (seems even a good practice to me to always annotate function types)

Will do. Can you give an example?

> why not use a function instead of a variable assignation for a function declaration?

What do you mean? Example?

> I would simplify with return doNotTrack === '1' || doNotTrack === 'yes', but it's really really minor

I prefer if statements for code readability. Not a big fan of || and ternary code for example but it's very subjective. If it's hurting the eyes too much for the team, I can change.

achauve: > > add return type hint as the function is exported (seems even a good practice to me to always annotate function types)
> 
> Will do. Can you give an example?
> 

Sorry, yes of course.

```typescript
let sum = (x: number, y: number): number => {
    return x + y;
}
```

In your case, adding boolean:
```typescript
export const doNotTrackSettingFromNavigatorIsActivated = (): boolean => {
...
}
```

> > why not use a function instead of a variable assignation for a function declaration?
> 
> What do you mean? Example?

I mean using:

```typescript
export function doNotTrackSettingFromNavigatorIsActivated(): boolean {
}
```

instead of new let/const and arrow function syntax.

> 
> > I would simplify with return doNotTrack === '1' || doNotTrack === 'yes', but it's really really minor
> 
> I prefer if statements for code readability. Not a big fan of || and ternary code for example but it's very subjective. If it's hurting the eyes too much for the team, I can change.

In either proposition `||` is used, it was just about dropping some code (but no other modification).
Full example:

```typescript
export function doNotTrackSettingFromNavigatorIsActivated(): boolean { 
    const doNotTrack = window.doNotTrack || navigator.doNotTrack;
    return doNotTrack === '1' || doNotTrack === 'yes';
}
```

lutangar (approved):
Thanks @felix-lambert looks good. I'm about to squash commits and merge it, there are too many duplicated commits to keep history as is.

lutangar (changes requested):
Sorry @felix-lambert I spoke too fast... there are now conflcts with the base branch. 
Could you rebase?

lutangar:
Thanks! I notice just one last little thing, the commit message doesn't follow our conventions. 
> see https://github.com/dis-moi/extension/blob/master/docs/CONTRIBUTING.md#commits

In this case the commit message should be something like:
```
feat: respect do not track preference
```
> mind the prefix

lutangar (approved):
:+1:

PR #760, opened 2020-12-28
Display Youtube player in notice when youtube link found

gregoirelacoste:
I added on the Notice component the possibility of recognizing youtube URLs and replacing them with the player of the corresponding video. For that I have:
- created a getMessageWithYoutubePlayer file which takes the initial children (string) from the "Message" component and transforms it via a Regex by adding the youtube iframe, or returns nothing if the regex does not match
- imported this function into the "Message" component
- added a defautlMessageWithYoutubeVideo in generateNotice
- everything is visible in the storybook: http://localhost:6007/?path=/story/extension-notice-details--with-youtube-video

### DIFF
--- a/src/app/content/App/Notice/Details/Details.stories.tsx
+++ b/src/app/content/App/Notice/Details/Details.stories.tsx
@@ -6,7 +6,8 @@ import { withKnobs, text, date, number, boolean } from '@storybook/addon-knobs';
 import Faker from 'faker';
 import {
   defaultMessage,
-  generateStatefulNotice
+  generateStatefulNotice,
+  messageWithYoutubeVideo
 } from 'test/fakers/generateNotice';
 import { Details } from '.';
 import { subMonths } from 'date-fns';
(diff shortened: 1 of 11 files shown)


JalilArfaoui:
@gregoirelacoste This is why it does not work : https://stackoverflow.com/questions/36337086/my-youtube-video-wont-show-in-iframe/36337328

You need to change any https://www.youtube.com/watch?v=XXX url to https://www.youtube.com/embed/XXX in the iframe

gregoirelacoste:
@JalilArfaoui I just tested that's it

lutangar:
Hey @gregoirelacoste sorry for the late review!

I would rather keep `Message` as a *dumb* component and move the call to `getMessageWithYoutubePlayer` in `NoticeDetails` component. 
What do you think @JalilArfaoui ? It would also mean to add the call to `NoticePreview` as well...

Also to expect a validation from the UI/Product team (maybe from @newick or @MaartenLMEM) could you provide a screenshot of the result?

Thanks.

lutangar:
@JalilArfaoui can we consider that @gregoirelacoste won't fix these little issues?

gregoirelacoste:
hi @lutangar sorry for the late answer, please find here the screenshot of the feature : 
image

lutangar:
Thanks @gregoirelacoste ! @newick @MaartenLMEM could you review the UI of this feature?

newick:
It seems to work perfectly @lutangar !
thx @gregoirelacoste 🎁

lutangar:
@newick I noticed the :-1: is little bit off could you fix this?
I just reassigned this PR to you.

lutangar:
@JalilArfaoui re-review plz I fixed a few things

christpet:
Looks like this is all set right @lutangar @JalilArfaoui ? If yes let's merge!

lutangar:
I suppose this isn't critical, but I'm the one who finished  the job, so mightn't be the best judge, and no one else gave a review on since then... I'm adding @gregoirelacoste as a reviewer!

lutangar:
@gregoirelacoste can't add you as a reviewer since you were the original author... let's merge this and move on...

Issue #360, opened 2019-09-05
Could we differentiate errors 400 from 500 ?


JalilArfaoui:
Could we differentiate 400 from 500?
Also 'Bad response from server' looks more like a 5xx rather than 4xx

_Originally posted by @lutangar in https://github.com/insitu-project/recommendations-webextension/pull/359_

PR #762, opened 2021-01-13
chore: Differentiate 4xx Http errors from an actual server error

lutangar:
I tried not to impact the current behavior but just to change the 4xx error behavior.

> see #360 for history

### DIFF
--- a/src/libs/api/APIStatusCodeError.ts
+++ b/src/libs/api/APIStatusCodeError.ts
@@ -3,7 +3,7 @@ export class APIStatusCodeError extends Error {
   public readonly statusCode: number;
 
   constructor(response: Response) {
-    super('Bad response from server');
+    super(response.statusText || 'Bad response from server');
     this.name = 'APIStatusCodeError';
     this.response = response;
     this.statusCode = response.status;
--- a/src/libs/api/call.ts
+++ b/src/libs/api/call.ts
@@ -27,16 +27,20 @@ export const buildQueryString: BuildQueryString = R.ifElse(
   )
 );
 
+export const checkStatus = (response: Response) => {
+  const { status } = response;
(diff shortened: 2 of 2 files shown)


lutangar:
ping @JalilArfaoui

lutangar:
ping @JalilArfaoui

src/api/call.ts
```diff
@@ -47,12 +58,11 @@ export const post = (path: string, data: {} | []) =>
...
-  }).then(response => {
-    if (response.status >= 400) {
-      throw new Error('Bad response from server');
-    }
-    if (response.status === 204) {
-      return true;
-    }
-    return response.json();
-  });
+  })
```
JalilArfaoui, line 61: Why only on `post` calls ?

src/api/call.ts
```diff
@@ -39,6 +39,17 @@ export const get = (path: string, data: object = {}) => {
   );
 };
 
+export const checkStatus = (response: Response) => {
+  const { status } = response;
+  if (status >= 200 && status < 300) {
+    return response;
+  } else if (status >= 500) {
+    throw new Error('Bad response from server');
```
JalilArfaoui, line 47: Maybe we could use `APIStatusCodeError` ?

lutangar: Sure, I updated `APIStatusCodeError` to use statusText as well.

src/api/call.ts
lutangar: Behavior is now the same for both.

JalilArfaoui (approved):

JalilArfaoui:
@lutangar so sorry for my review delay !

JalilArfaoui:
This is conflicting new

PR #817, opened 2021-03-04
chore: Fix licence format in package.json

lutangar:
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license
> https://spdx.org/licenses/


gregoirelacoste (approved):
Syntax is ok

AGPL-3.0-or-later (GNU Affero General Public License v3.0 or later) is a free / libre OSI approved label
In summary : _Everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed._ 

Details: @https://spdx.org/licenses/AGPL-3.0-or-later.html 

looks good for me

PR #845, opened 2021-03-16
build: handle firefox development

felix-lambert:


README.md
```diff
@@ -78,6 +80,15 @@ yarn start
 > Chrom(e|ium) browser is recommend for development.
 > see https://developer.chrome.com/extensions/getstarted#manifest
 
+### For firefox development extension
+
+```bash
+npm install --global web-ext
```
lutangar, line 86: This part is not required since `web-ext` is already a dependency.

package.json
```diff
@@ -9,6 +9,7 @@
     "start": "webpack --watch --progress --mode=development",
     "start:profiles": "webpack-dev-server --progress --mode=development --config webpack.profiles.config.js --inline",
     "start:firefox": "webpack --watch --progress --mode=development --env.PLATFORM=firefox",
+    "run:firefox": "cd build/development/firefox && web-ext run",
```
lutangar, line 12: Prefer to run it via `start:firefox` in dev, since it uses webpack watch mode.

README.md
```diff
@@ -78,6 +80,15 @@ yarn start
...
+yarn start:firefox
+yarn run:firefox
```
lutangar, line 88:
```
yarn start:firefox
```
is sufficient.

lutangar (changes requested):

package.json
felix-lambert: It's a command that only runs the navigator with the extension installed. You think we don't need this command?

lutangar: Oh I see! This is actually interesting let's keep it.

felix-lambert: Will check if webpack watch mode works with this command

lutangar (approved):

PR #853, opened 2021-03-18
English translation - extension

christpet:
This is good to go!

### DIFF
--- a/src/i18n/resources/en/extension.json
+++ b/src/i18n/resources/en/extension.json
@@ -24,7 +24,7 @@
   },
   "date": {
     "short": "{{date, MM/DD/YYYY}}",
-    "medium": "Le {{date, MM/DD/YYYY}}"
+    "medium": "{{date, MM/DD/YYYY}}"
   },
   "menu": {
     "my_account": "My account",
--- a/src/i18n/resources/en/extension.json
+++ b/src/i18n/resources/en/extension.json
@@ -1,110 +1,110 @@
 {
   "common": {
-    "or": "ou",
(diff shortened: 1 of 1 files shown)


src/i18n/resources/en/extension.json
```diff
@@ -1,110 +1,110 @@
...
+      "back": "back",
+      "messages": "Back to messages",
+      "submissions": "Back to contributions"
     },
-    "manage_subscription": "Gérer mes éclaireurs"
+    "manage_subscription": "Manage my guides"
   },
   "date": {
-    "short": "{{date, DD/MM/YYYY}}",
-    "medium": "Le {{date, DD/MM/YYYY}}"
```
gregoirelacoste, line 27: We must remove the "Le"

gregoirelacoste (changes requested):

src/i18n/resources/en/extension.json
christpet: Ah good find ;)

src/i18n/resources/en/extension.json
```diff
@@ -1,110 +1,110 @@
...
+    "short": "{{date, MM/DD/YYYY}}",
+    "medium": "Le {{date, MM/DD/YYYY}}"
```
christpet, line 27:
```suggestion
    "medium": "{{date, MM/DD/YYYY}}"
```

lutangar:
> https://github.com/dis-moi/extension/pull/854#issuecomment-803881548

PR #854, opened 2021-03-18
English translation - profiles

christpet:
All set - the only thing I'm not sure about is the changed date format.

### DIFF
--- a/src/i18n/resources/en/profiles.json
+++ b/src/i18n/resources/en/profiles.json
@@ -1,77 +1,77 @@
 {
   "common": {
-    "sources": "Les sources",
-    "subscriber": "Abonné(e)",
-    "similar_profiles": "Profils similaires",
-    "no_more_contribution": "Pas d'autres contributions",
+    "sources": "Guides",
+    "subscriber": "Subscribers",
+    "similar_profiles": "Similar profiles",
+    "no_more_contribution": "No more contributions",
     "all": "Tous"
   },
   "action": {
-    "see_contributions": "Voir ses contributions",
-    "add_dismoi_to_browser": "Ajouter Dismoi à mon navigateur",
-    "add_to_my_browser": "Ajouter à mon navigateur ",
-    "follow": "Suivre",
-    "know_more_dismoi": "En savoir plus sur <1>DisMoi</1> et ses sources",
-    "know_more": "En savoir plus",
-    "stop_following": "Ne plus suivre",
-    "see_more": "Voir plus",
-    "see_all": "Voir tout",
-    "see_more_details": "Plus de détails",
-    "see_context": "Voir en contexte",
-    "filter_by": "Filtrer par :"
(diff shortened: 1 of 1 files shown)


src/i18n/resources/en/profiles.json
```diff
@@ -1,77 +1,77 @@
...
   "common": {
-    "sources": "Les sources",
-    "subscriber": "Abonné(e)",
-    "similar_profiles": "Profils similaires",
-    "no_more_contribution": "Pas d'autres contributions",
+    "sources": "Guides",
+    "subscriber": "Subscribers",
+    "similar_profiles": "Similar profiles",
+    "no_more_contribution": "No more contributions",
     "all": "Tous"
```
gregoirelacoste, line 7: it miss "all" ?

src/i18n/resources/en/profiles.json
```diff
@@ -1,77 +1,77 @@
...
+      "description_dismoi":"DisMoi lets internet users, media and experts share helpful information directly on the webpages you visit."
     }
   },
   "notice": {
-    "screenshot_alt": "Rendu de la contribution sur {{exampleMatchingUrl}} une fois l'extension installée.",
-    "pined_on": "<0>Message épinglé sur <1>{{exampleMatchingUrl}}</1></0> et d'autres pages web",
-    "since": "Visible depuis le {{date, DD/MM/YYYY}}"
+    "screenshot_alt": "Shows the contribution on {{exampleMatchingUrl}} once the extension is installed.",
+    "pined_on": "<0>Message pinned on <1>{{exampleMatchingUrl}}</1></0> and other webpages",
+    "since": "Since {{date, MM/DD/YYYY}}"
```
gregoirelacoste, line 75: Fine !

gregoirelacoste (changes requested):
Nice, All that's missing is "all", thank you !!

src/i18n/resources/en/profiles.json
christpet: Ah yes I did!

christpet: Thanks!

christpet, line 7:
```suggestion
    "all": "All"
```

lutangar:
@gregoirelacoste  you let unformatted commit messages slip through the pipeline.
I might add an "automated thing" to prevent this to happens unnoticed.

> For reference, this convention is responsible for setting  the extension version number and generates changelog on new releases:
> https://github.com/dis-moi/extension/releases
>
> This is mentionned here and there:
> https://github.com/dis-moi/extension/blob/master/docs/CONTRIBUTING.md#commits
> https://github.com/dis-moi/docs/pull/1/files

@christpet I'd be more confortable if a tech person would hit the merge button as long as we can let human errors go through to the pipe unnoticed.

PR #864, opened 2021-03-23
fix(Avatar): adds height value

newick:
Quick corrects #766


christpet (approved):
Looks great, thank you @newick !

PR #875, opened 2021-03-25
fix(NoticesList): way to display more than 2 notices

### LINKED ISSUE #850: If there are more than 2 notices on a page, they are invisible
Let see that there are more than 2 notices on a same page and give access to them

**Target user**
DisMoi addon users

**Fill in the following areas if you have the time...**

**User stories**
How will the user experience this feature/improvement?
1. The visit a page where there a 3 (or more) notices matching
2. User can see the two first preview, and I can see that there are others 

**Technical implementation notes**
Display all bubbles and add a scrollbar

**Additional context**
It s also an issue for a contributor whose contributions are invisible, which is frustrating (got a mail about this case (see [this page](https://www.sephora.fr/shop/soin-visage/demaquillant-et-nettoyant-c7633/)) from Ethik'do wher bubble 1524 is not visible.
image

newick:
A way to fix #850

@MaartenLMEM / @christpet visually it would be something like 
image

@gregoirelacoste could you find a way to hide the message if there's less than 2 notices?

### DIFF
--- a/src/app/content/App/Notice/List/List.stories.tsx
+++ b/src/app/content/App/Notice/List/List.stories.tsx
@@ -12,6 +12,7 @@ import { ListScreen } from '.';
 
 const firstMessage = defaultMessage;
 const secondMessage = `De nombreux clients mécontents de Pixmania et ses vendeurs s'expriment sur les réseaux sociaux depuis 2016. Les plaintes continuent en 2017, 2018 et encore en 2019 si l'on se réfère au forum Que Choisir.`;
+const thirdMessage = 'yolo';
 const commonProps = {
   close: action('close'),
   dismiss: action('dismiss'),
@@ -51,6 +52,25 @@ storiesOf('Extension/Notice/List', module)
         generateStatefulNotice({
           dismissed: boolean('dismissed(2)', false, 'second'),
           message: `<p>${text('message(2)', secondMessage, 'second')}</p>`
+        })
+      ]}
+    />
+  ))
+  .add('3 notices', () => (
+    <ListScreen
+      {...commonProps}
+      notices={[
+        generateStatefulNotice({
+          dismissed: boolean('dismissed(1)', false, 'first'),
+          message: `<p>${text('message(1)', firstMessage, 'first')}</p>`
+        }),
+        generateStatefulNotice({
+          dismissed: boolean('dismissed(2)', false, 'second'),
+          message: `<p>${text('message(2)', secondMessage, 'second')}</p>`
+        }),
+        generateStatefulNotice({
+          dismissed: boolean('dismissed(3)', false, 'third'),
+          message: `<p>${text('message(3)', thirdMessage, 'third')}</p>`
         })
       ]}
     />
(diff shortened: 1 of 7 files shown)


MaartenLMEM:
**Wording** : use `contributions` instead of bulles, which is more internal.
**Display** : Is it possible to show the 1st head of the third notice instead of this "there is more" line ?
Because currenly, button "Poster" collide with any label right above.

newick:
@MaartenLMEM something like this?

image

MaartenLMEM:
I prefer this direction yes, seems me more clear. What do you think @christpet ?

Oh ! Actuallly what about puting the button "Poster" after the last bubble into the scroll area ? Isn't it more consistent with for contribution process on this screen (I read what is written here and I add something if I want) ?

newick:
> Oh ! Actuallly what about puting the button "Poster" after the last bubble into the scroll area ? Isn't it more consistent with for contribution process on this screen (I read what is written here and I add something if I want) ?

Depends of what we want.
If it's to get contributions, the button should be displayed all the time.
If we want Users to be sure that they won't publish duplicate content, the button could be in the scrollbar. 

I think the last solution works only if the Users can share a contribution.

@prk-dismoi, do you have any idea about that?

gregoirelacoste:
>  the button "Poster" after the last bubble into the scroll area ?
It might be discouraging to scroll to post if there is too much content ?
@newick what can i do to help you on it ?

christpet:
> @christpet so without number of contributions ? even with something like this ?
> image

Good thinking, but I'm not sure there's a clear need for that yet. I always try to operate with the principle of removing as much as possible from the interface unless it's necessary. We can add it later if needed!

gregoirelacoste:
> I always try to operate with the principle of removing as much as possible from the interface unless it's necessary

ok !

gregoirelacoste:
result for last push: 
image

MaartenLMEM:
ok for our contribution partners, I think for users also :)

gregoirelacoste:
> ok for our contribution partners, I think for users also :)

- if we favor users > no visible "post" button
- if we favor contributor > visible "post" button
- 
who to favor?

MaartenLMEM:
@christpet choosed "post button" - Follow him !
(By the way I agree because it is the most simple)

newick:
@gregoirelacoste give me 15min to try the box shadow

newick:
first try is ugly 
image

I try something else

christpet:
@newick almost there - make the shadow go the other direction!

newick:
@christpet not sure it helps 
image

gregoirelacoste:
> @christpet not sure it helps
> image

would that be fine to add a wrapper to "poster" button (100%width) and  give a box shadow only on top ?

newick:
good idea @gregoirelacoste 
image

newick:
@gregoirelacoste is there a way to use a prop to display the box-shadow only for more than 2 contribs?

gregoirelacoste:
@newick yes, i do that at 2pm if you want

christpet:
@newick @gregoirelacoste looks great. Yes if you can do that quickly at 2pm then let's ship it!

gregoirelacoste:
last push : 
for 3 and more contributions
image

for less
image

christpet (approved):
Excellent work @gregoirelacoste and @newick ! Looks great to me. This looks like it is good to go. Shall we merge?

gregoirelacoste:
@christpet i do

PR #884, opened 2021-03-29
fix(newOnboarding): video link && wording

gregoirelacoste:


lutangar (approved):

gregoirelacoste:
if you have something to notice on new onboarding please do here @christpet @newick

PR #885, opened 2021-03-29
fix(newOnboarding): video muted

gregoirelacoste:


gregoirelacoste:
@newick @christpet could you make a last review from front and tell if it's ok before last publish ?

newick (approved):

PR #895, opened 2021-04-12
fix(routerEn): ProfileTabs links

gregoirelacoste:


yarn.lock
```diff
@@ -1777,7 +1777,7 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.10.2", "@babel/runtime@^7.12.5", "@babel/runtime@^7.13.10", "@babel/runtime@^7.8.7":
+"@babel/runtime@^7.10.2", "@babel/runtime@^7.12.1", "@babel/runtime@^7.12.5", "@babel/runtime@^7.13.10", "@babel/runtime@^7.8.7":
```
JalilArfaoui, line 1780: A reason why you update that ?

JalilArfaoui (approved):

yarn.lock
gregoirelacoste: i have diff from an another branch

PR #900, opened 2021-04-13
Chore/eslint rules

lutangar:
Requesting comments on 2 unwritten rules we had on the codebase.

I got tired of mentioning it on PR, we probably should have a rule enforcing AND fixing this *automagically* for us instead.

## Import order
I just went for the default rule as a starter:
> https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md#importorder-enforce-a-convention-in-module-import-order

This is a fixable rule so it'll get fixed with a `eslint --fix`.
I added `prettier-eslint` for that, running prettier is now fixing `eslint` fixable errors.

## Relative imports
I tried to limit relative import depth to something sensible using `no-restricted-imports` but could make the *ignore* `patterns` work.

I think that `.gitignore` syntax just does make no sense at all in such context.
```
"no-restricted-imports": ["error", {
  "patterns": ["../../../**", "!../../../**/package.json"]
}],
```

> __Resources__
> - https://eslint.org/docs/rules/no-restricted-imports#options
> - https://github.com/kaelzhang/node-ignore
> - https://github.com/eslint/eslint/blob/master/lib/rules/no-restricted-imports.js#L120

The idea was to gently push toward using alias (or other path resolver) instead of relative import.

### Example
For example, instead of `import { Button } '../../../components'` prefer `import { Button } 'components';`.

But this might solves itself with better tooling later. Ideas?

## More
Is there other rules that should land in this PR as well?
Are there any other things you're doing manually which you're finding tiresome?

### DIFF
--- a/src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
+++ b/src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
@@ -1,12 +1,12 @@
 import React, { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Field, InjectedFormProps } from 'redux-form';
-import withReduxForm from './withReduxForm';
 import { Form } from 'components/atoms/Forms';
 import { CenterContainer, BackgroundButton } from 'components/atoms';
 import { InputField, TextareaField } from 'components/organisms/Fields';
 import FormErrors from 'components/molecules/FormErrors';
 import { Contribution } from 'app/lmem/notice';
+import withReduxForm from './withReduxForm';
 
 export interface SubmitContributionFormOwnProps {
   onUrlChange: (url: string) => void;
--- a/src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
+++ b/src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
@@ -1,12 +1,12 @@
 import React, { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Field, InjectedFormProps } from 'redux-form';
+import withReduxForm from './withReduxForm';
 import { Form } from 'components/atoms/Forms';
 import { CenterContainer, BackgroundButton } from 'components/atoms';
 import { InputField, TextareaField } from 'components/organisms/Fields';
 import FormErrors from 'components/molecules/FormErrors';
 import { Contribution } from 'app/lmem/notice';
-import withReduxForm from './withReduxForm';
 
 export interface SubmitContributionFormOwnProps {
   onUrlChange: (url: string) => void;
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -12,6 +12,12 @@
   "settings": {
     "react": {
       "version": "detect" // React version. "detect" automatically picks the version you have installed.
+    },
+    "import/resolver": {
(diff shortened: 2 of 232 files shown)


src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
```diff
@@ -1,12 +1,12 @@
 import React, { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Field, InjectedFormProps } from 'redux-form';
+import withReduxForm from './withReduxForm';
```
JalilArfaoui, line 4: It does not seem to work … relative imports should be after absolute ones (https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md)

JalilArfaoui (changes requested):

src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
lutangar: I think we just need to tweak the rule a bit.

So I guess your fine with this setting? No changes?
```
// 1. node "builtin" modules
import fs from 'fs';
import path from 'path';
// 2. "external" modules
import _ from 'lodash';
import chalk from 'chalk';
// 3. "internal" modules
// (if you have configured your path or webpack to handle your internal paths differently)
import foo from 'src/foo';
// 4. modules from a "parent" directory
import foo from '../foo';
import qux from '../../foo/qux';
// 5. "sibling" modules from the same or a sibling's directory
import bar from './bar';
import baz from './bar/baz';
// 6. "index" of the current directory
import main from './';
// 7. "object"-imports (only available in TypeScript)
import log = console.log;
```
@gregoirelacoste comments?

JalilArfaoui: Yes (default one, right ?) it’s what I was doing anyway …

I’d like to have the `4.` be more specific with outer folders first, but I’m tinkering

Now, we just have to make it work like expected :-)

lutangar:
@JalilArfaoui while I'm at it, I was kind of stuck trying to implement a rule for relative imports, I thought you might have some suggestions as well.

src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
gregoirelacoste: @lutangar Nice! I usually import react and its modules always first, we can immediately see if we are in a component, it might be good to put it first when possible?

gregoirelacoste (approved):
LGTM

newick (commented):
I can't say anything about the code but I like the idea!

src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/SubmitContributionForm.tsx
lutangar: > import react and its modules always first,

Not sure I can enforce this with this rule, but if put first the react import will stay first.

@JalilArfaoui the config is now working as expected

lutangar: ping @JalilArfaoui

JalilArfaoui (approved):
Looks nice now !

JalilArfaoui:
Needs a rebase, though

Issue #904, opened 2021-04-16
Bug staging : post a contribution generates infinite loader


MaartenLMEM:
Related to new feature in staging https://github.com/dis-moi/extension/pull/800

Firefox Staging 3.71.0 and Proding 3.71.00
Reproduced 4 times with désactiver /activer and then uninstall / install between each try

**bug** 
I try to publish a contribution, but after clicking on validate the button stays in loading state.

To reproduce
1. Clic on D icon one any page
2. Fill name, email and message
3. Clic on preview
4. Clic on validate
Bug happens at step 4
![image](https://user-images.githubusercontent.com/7824485/115051695-64246580-9edd-11eb-97d5-524eba44665d.png)

MaartenLMEM:
Background random screenshot (where it's red ^^)
![image](https://user-images.githubusercontent.com/7824485/115051444-1f003380-9edd-11eb-80cf-7832de04d68d.png)

MaartenLMEM:
update : Christian found same bug and described it directly in the feature https://github.com/dis-moi/extension/pull/800

PR #800, opened 2021-02-23
feat: Post contribution and question directly to the backend

lutangar:
This is the first draft allowing to post `Contribution` and other `Question` to the backend.
No more code sending email out there in the wild :tiger: nor more *sendinblue*.

Form submission action is sent to the background and then to the backend API to avoid CORS issues.

This actually removes a lot of code :sweat_smile: 

> EDIT: The hidden goal here is to gain back automated publication process and to ship more :shipit: 

> This PR works with https://github.com/dis-moi/backend/pull/161

PR #912, opened 2021-04-19
Fix/submit contribution firefox

lutangar:
> see details in commit bodies

### DIFF
--- a/src/app/actions/i18n.ts
+++ b/src/app/actions/i18n.ts
@@ -1,0 +1,10 @@
+import { BaseAction } from '.';
+
+export const I18N_READY = 'EXTENSION/I18N_READY';
+export interface I18nReady extends BaseAction {
+  type: typeof I18N_READY;
+}
+
+export const i18nReady = (): I18nReady => ({
+  type: I18N_READY
+});
--- a/src/app/store/stripFormMeta.ts
+++ b/src/app/store/stripFormMeta.ts
@@ -1,0 +1,9 @@
+import * as R from 'ramda';
+import { AppAction } from '../actions';
+
+const stripFormMeta = R.compose(
+  R.dissocPath<AppAction>(['meta', 'resolve']),
+  R.dissocPath<AppAction>(['meta', 'reject'])
+);
+
+export default stripFormMeta;
--- a/src/webext/sendMessageToBackground.ts
+++ b/src/webext/sendMessageToBackground.ts
@@ -1,4 +1,8 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+import { serialize } from './serializer';
+
 const sendMessageToBackground = (message: unknown) =>
-  browser.runtime.sendMessage(message);
+  // @ts-ignore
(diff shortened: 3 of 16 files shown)


src/webext/sendMessageToBackground.ts
```diff
@@ -1,4 +1,8 @@
+/* eslint-disable @typescript-eslint/ban-ts-ignore */
+import { serialize } from './serializer';
+
 const sendMessageToBackground = (message: unknown) =>
-  browser.runtime.sendMessage(message);
+  // @ts-ignore
+  browser.runtime.sendMessage(serialize(message));
```
JalilArfaoui, line 6: Serialization is done automatically if a serializable object is passed … 
https://developer.mozilla.org/fr/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage
Can you explain why we’d do this manually ?

src/app/store/stripFormMeta.ts
```diff
@@ -0,0 +1,9 @@
+import * as R from 'ramda';
+import { AppAction } from '../actions';
+
+const stripFormMeta = R.compose(
```
JalilArfaoui, line 4: Nice …but you don’t use is ?

JalilArfaoui (commented):
I guess maybe you forget to use stripFormMeta ? Then we should not need to serialize/deserialize manually … ?

src/webext/sendMessageToBackground.ts
lutangar: See b6fa7ad body and #904

src/app/store/stripFormMeta.ts
lutangar: That was my first attempt but this is a bit limited and this is causing other issue, since we're expecting these metas to be defined in other sagas.
Serialization is a more radical solution for other potential issues and keep action consistent across content/background. I thought it would also fix the empty error message but unfortunately it doesn't. 
>  ;) see https://github.com/dis-moi/extension/pull/800#discussion_r585066876

MaartenLMEM:
Can this be soon merged ? @lutangar @JalilArfaoui

src/app/actions/i18n.ts
```diff
@@ -0,0 +1,10 @@
+import { BaseAction } from '.';
+
+export const I18N_READY = 'EXTENSION/I18N_READY';
+export interface I18nReady extends BaseAction {
+  type: typeof I18N_READY;
+}
+
+export const i18nReady = (): I18nReady => ({
```
gregoirelacoste, line 8: what does this action do?

lutangar: in fact it does not do anything right now!

lutangar: I mean we dispatch it but we don't do anything else with it

gregoirelacoste (approved):

MaartenLMEM:
Can I merge it @gregoirelacoste ? One review is enough here isn't it ?

gregoirelacoste:
@MaartenLMEM yes!

PR #915, opened 2021-04-24
fix: extension completely hangs after 6|10 retries 

### LINKED ISSUE #861: Chrome extension blocked on Chloé and Maarten's computers and probably others
**Describe the bug**
On Chloé's computer on every page, DisMoi is grey : no notice, no interaction possible. 
It's possible she had same problem on other computer, but not sure.

**Vidéo**
This Vidéo shows bug but also informations from extension background.
It has been recorded during our call : https://drive.google.com/file/d/1kqcqnqXR5raQYVGhpUAObLNK6Dc7tUTC/view

**To Reproduce**
I don't know

**Desktop):**
 - OS: Windows
 - Browser : Chrome 
 - Version : 89 0.4389 90

**Extension**
- Version : 3.61
- Installed from store
- Installation date : don't know, probably recently
It's a new computer she just bought, refurbed.

JalilArfaoui:
fix: extension completely hangs after 6|10 retries of fetching contributors|matching-contexts

fixes #861

### DIFF
--- a/src/app/sagas/effects/callAndRetry.ts
+++ b/src/app/sagas/effects/callAndRetry.ts
@@ -18,12 +18,15 @@ const defaultOptions: Options = {
   maximumRetryDelayInMinutes: 60 * 24
 };
 
-export const createCallAndRetry = (options: Partial<Options>) =>
+export const createCallAndRetry = (givenOptions: Partial<Options>) =>
   function* callAndRetry<Fn extends AnyFunction>(
     fn: Fn,
     ...args: Parameters<Fn>
   ): SagaIterator {
-    const o: Options = R.mergeRight(defaultOptions, options) as Options;
+    const options: Options = R.mergeRight(
+      defaultOptions,
+      givenOptions
+    ) as Options;
 
     function* attempt(
       attemptNumber: number,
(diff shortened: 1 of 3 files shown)


src/app/sagas/effects/callAndRetry.ts
```diff
@@ -36,7 +35,7 @@ export const createCallAndRetry = (options: Partial<Options>) =>
         if ('onError' in o && o.onError) {
           yield call(o.onError, e, attemptNumber);
         }
-        if (attemptNumber < o.maximumAttempts) {
+        if (!o.maximumAttempts || attemptNumber < o.maximumAttempts) {
```
felix-lambert, line 38: What us o referring to? Maybe we could be more explicit with the naming?

felix-lambert (commented):
Just a review based on naming. It's of course a non blocking one

src/app/sagas/effects/callAndRetry.ts
JalilArfaoui: You’re right, I’ve renamed it to `options`

felix-lambert (approved):

lutangar (approved):

lutangar:
I thought there was an unhandled `exception` left somewhere when the backend wasn't reachable. Does it solves this issue and catch this one? In `dev` this was easy to test, just dont start the backend and the extension crash nastily.

JalilArfaoui:
@lutangar we were trying up to 10 times … Now with this, we’ll keep trying until network or backend recovers

PR #917, opened 2021-04-25
perf: change matching contexts periodical refresh logic

JalilArfaoui:
This is in order to avoid parallel fetching attemps in case previous one still hasn’t succeeded. With this, we only request a new refresh once previous one has succeeded.

Relates to #861


lutangar (approved):
> we only request a new refresh once previous one has succeeded.

Or when the previous one has failed? This is probably the case but wanted a confirmation on this.

JalilArfaoui:
@lutangar :  
- If it fails, we keep trying
- but only when it succeeded, we plan the next trial (via `delay`)

PR #919, opened 2021-04-25
perf: change contributors periodical refresh logic

JalilArfaoui:
This is in order to avoid parallel fetching attemps in case previous one still hasn’t succeeded. With this, we only request a new refresh once previous one has succeeded.

Same spirit as #917 but for contributors

Related to #861


lutangar (approved):
> See question in https://github.com/dis-moi/extension/pull/917#pullrequestreview-644452855

When I see our implementation I wonder if we missed something because it seems we're not using the cancel APIs, neither on the fetch side nor on the saga side. But It might not be relevant here... :thinking:

felix-lambert (approved):

JalilArfaoui:
Actually, our `callAndRetry` function does use it behind the scene, along with all `takeLatest` effects … but there is definetely room for improvement … but I prefer not trigger useless network requests, than cancelling them …

Issue #801, opened 2021-02-24
Error messages when 403, etc. happen


MaartenLMEM:
For extensions UI sand contributor UIs
Edge cases like 403 (connexion off)

christpet:
To clarify... Is this issue asking for a design to be created for those edge cases? Or do the designs already exist and they need to be implemented? @MaartenLMEM

MaartenLMEM:
Yes UI designs already exist. But, If I remember correctly,  they are quite general. We did not make a list of kind of errors, on which screens, it s just general. Neverthelss, I think I would be an improvment, even if it doesn't precisely fit to each user situation. It's better than no feedback.
@coffeeandpixel knows more about this matter then I do. He can provide 1st designs, etc.

christpet:
Okay thanks, this could be a relatively easy one to implement. I'll put it in the "to do" column for anyone to take when they have a moment.

In the meantime, @coffeeandpixel could you point us to those designs?

lutangar:
I agree with @JalilArfaoui this should be planned. I'm seeing this has been moved to the [Performance board](https://github.com/orgs/dis-moi/projects/12) but should we move this to the extension todo as well @MaartenLMEM ?

MaartenLMEM:
I agree, I am just waiting for the design.

PR #923, opened 2021-04-26
Badge on network error

JalilArfaoui:
Regarding #801, despite the fact that we are waiting for design to implement, I wanted to propose this POC of a simple visual feedback when network is unavailable or backend does not respond rightfully. 

I made a little badge :  
image

That has a little tooltip : 
image

It’s totally non-final, but it aims to open the discussion to finally have some feedback !

### DIFF
--- a/src/app/actions/index.ts
+++ b/src/app/actions/index.ts
@@ -25,6 +25,7 @@ import {
 } from './ui';
 import {
   FeedbackOnNoticeAction,
+  FetchNoticeFailureAction,
   MarkNoticeReadAction,
   NoNoticesDisplayedAction,
   NoticeDisplayedAction,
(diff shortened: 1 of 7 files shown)


lutangar (commented):
This is interesting :+1: 
I wonder if this could be more generic on how we want to handle error scenarios #801
An error occurred and we are...
- on a badge, how to show error here
- on the list of notices
- ...

I'd say badge is just another context where we should be able to show an error.

Not much not say about the technical implementation.
Maybe instead of `link` I'd use `connection` and use similar vocabularies as what exists for connection and "not internet" errors.

newick:
Good idea!

I could find a wording issue because `backend` is a very technical word ;)

gregoirelacoste (commented):
Good idea, 
we should also think of a clearer message, displayed when opening the extension that would explain what to do, depending on the problem (reload the page etc)

JalilArfaoui:
@MaartenLMEM  I see you closed this one … Can you confirm we won’t need this feature ?

MaartenLMEM:
sorry was a mistake

PR #959, opened 2021-05-10
fix(extension): initialize i18next before store

### LINKED ISSUE #950: staging : screen titles have disappreared in contribution & question flow
**Describe the bug**
Lorsque je contribue ou lorsque je demande une alternative, les titres des différents écrans ont été remplacé par le logo.

- En production actuellement sur Firefox et Chrome : 
image

- En staging firefox 3.73.0 
image

=> Le titre de la version prod a été remplacé par le logo en staging

**To Reproduce**
1. install staging 3.73.0 or more recent version on firefox
2. Click on "demander une info" or "post"
3. See error

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser : Firefox

**Extension**
- Version staging 3.73.0
- Installed from store

gregoirelacoste:
on Content app, 18n have to be initialized before store to get the titles 
closes #950

### DIFF
--- a/src/app/content/index.ts
+++ b/src/app/content/index.ts
@@ -29,22 +29,28 @@ if (!(window as CustomWindow).__BULLES__CONTENT_SCRIPT_INJECTED__) {
       });
     }
 
-    const store = require('./store').default;
-    const documentReady = require('../utils/documentReady').default;
-    const externalClickHandler = require('./externalClickHandler').default;
-    const handleExternalClick = externalClickHandler(store);
+    const initApp = () => {
+      const store = require('./store').default;
+      const documentReady = require('../utils/documentReady').default;
+      const externalClickHandler = require('./externalClickHandler').default;
+      const handleExternalClick = externalClickHandler(store);
 
-    (window as CustomWindow).__BULLES__CONTENT_SCRIPT_INJECTED__ = true;
+      (window as CustomWindow).__BULLES__CONTENT_SCRIPT_INJECTED__ = true;
 
-    documentReady.then(() => {
-      document.addEventListener('click', handleExternalClick);
+      return documentReady.then(() => {
+        document.addEventListener('click', handleExternalClick);
 
-      window.addEventListener('unload', () => {
-        document.removeEventListener('click', handleExternalClick);
+        window.addEventListener('unload', () => {
+          document.removeEventListener('click', handleExternalClick);
+        });
       });
-    });
+    };
 
-    i18n.init(options).then(() => {});
+    i18n
+      .init(options)
+      .then()
+      .catch(e => captureException(e))
+      .finally(() => initApp()); // i18n have to be initialized before store to get the titles
   } catch (error) {
     captureException(error);
   }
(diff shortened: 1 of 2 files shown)


JalilArfaoui:
@gregoirelacoste this should work, but in order to get rid of the dependency between the store and i18n, we could instead keep the localization in the vue, and avoid having translated text in the store altogether. 

To do that, I suggest we remove `if (title) title = i18next.t(title);` from `withTitle` and change for example the `Notification` component to apply `t()` … 

This way, we avoid having a fragile app initialization sequence …

What do you think ?

gregoirelacoste:
> This way, we avoid having a fragile app initialization sequence …
Is it really fragile? Because `appInit()` is always initialized, with or without title, with promise.finally().

If we apply t everywhere, we have to change all components (at least 10), i think that is less maintainable and there is a risk to forget one... no?

lutangar:
I'm the one who asked @JalilArfaoui to get involved here, at first it was because I didn't like handling the async flow in the app index. I do prefer to handle all async flows with actions and then with saga, it give way more control.
(That was the reason I first introduce the `i18nReady` action a few weeks back)

> I suggest we remove if (title) title = i18next.t(title); from withTitle and change for example the Notification component to apply t()

I totally agree. Plus if it means we don't need to wait for the i18n to load that would be totally awesome!

gregoirelacoste:
> Plus if it means we don't need to wait for the i18n to load that would be totally awesome!

We need to detect the language with i18next, so in any case we have to wait for the lib at some point

gregoirelacoste:
I agree with the problem to init i18n before the app. In my opinion it makes sense to use WithTitle for the translation, it allows to centralize all the logic of the titles in the same hoc.
So I allow myself to suggest you another solution that works without initializing i18n before @JalilArfaoui @lutangar 
if you're not ok i will put t function inside the components

JalilArfaoui (approved):
It’s OK for me, no big deal.

But I still think this is not optimal for a few reasons : 
 - `withTitle` has more than one responsability : Allow for a component to set the frame title + translating this title.  To me, ideally `withTitle` should know nothing about translation.
 - We still store translated texts in the store, which seems to me like storing a «computed state» …
 
 You’re right about the risk of mistake in forgetting to `t()` some text in some component, but to avoid that we have TypeScript and branded types, we should differentiate  :
 
 ```ts
 type TextLocalizationKey = Brand<string, 'TextLocalizationKey'> // to type any translation key, that must not be renders as is
 type LocalizedText = Brand<string, 'LocalizedText'> // to type already translated texts
 ```
 
 With that, if a component waits for a translated text and is given a translation key, TypeScript will yell at us !

gregoirelacoste:
> ```ts
> type TextLocalizationKey = Brand<string, 'TextLocalizationKey'> // to type any translation key, that must not be renders as is
> type LocalizedText = Brand<string, 'LocalizedText'> // to type already translated texts
> ```

very nice solution! 
You approved the changes, it means that we merge this issue, and create a new one with low priority or refactor label ?

JalilArfaoui:
It means it’s up to you to decide :-)

If it feels it would take some time, and you think your actual solution fixes an actual issue in production, go ahead to merge and create a `refactor: move the translation out of withTitle` follow-up issue !

PR #971, opened 2021-05-13
Feat/example link on followed list

### LINKED ISSUE #966: Add link "voir ses contributions" on page "followed profiles"
Link "voir ses contributions" has disappeared on "contributeurs suivis" onglet 

image

gregoirelacoste:
closes #966


src/app/profiles/store/actions/popin.ts
```diff
@@ -0,0 +1,13 @@
+import { ContextPopinState } from '../reducers/popin.reducer';
+import store from '../index';
+
+export const SET_POPIN = 'SET_POPIN';
+
+export const setPopin = (payload: ContextPopinState) => ({
```
lutangar, line 6: Usually we avoid magic setter action, but I don't mind here.
However it should name accordingly an use the right interfaces.
```
export const setPopinState = (popinState: ContextPopinState): StandardAction => ({
...
```

Every action should be a  `StandardAction` see other actions for further references.

Also I don't know if it's always a `ContextPopin` or only a `Popin`, sometimes you're using and sometimes the other. You should be consistent.

src/app/profiles/store/reducers/popin.reducer.ts
```diff
@@ -0,0 +1,30 @@
+import { SET_POPIN, PopinActionReturn } from '../actions/popin';
+import { MouseEvent } from 'react';
+
+export interface ContextPopinState {
+  content: {
+    text: string;
+    btnLabel?: string;
+    onClick?: (params?: MouseEvent<HTMLButtonElement>) => void;
```
lutangar, line 8: It seem this onclick will alway apply to the btn.. mb your state would be better shaped smthing like:
```ts
{
    text: string; // or content
    btn?: ButtonDescriptor;
    opened: boolean;
}

interface ButtonDescriptor { 
      label?: string;
      onClick?: (params?: MouseEvent<HTMLButtonElement>) => void;
}
```

Your call.

src/app/profiles/store/reducers/index.ts
```diff
@@ -27,4 +29,5 @@ export interface ProfilesState extends RouterRootState {
   subscriptions: SubscriptionsCollectionState;
   notices: NoticesCollectionState;
   categories: CategoriesCollection;
+  popin: ContextPopinState;
```
lutangar, line 32:
```
popin: PopinState;
```
or
```
contextPopin: ContextPopinState;
```

lutangar (changes requested):
Quite a few things left to fix, since it's the first time you're implementing actions in this project that's quite normal. Feel free to give me a call.

You should also be careful with your import, prefer the "absolute" syntax.

src/app/profiles/store/selectors/contextPopin.ts
```diff
@@ -0,0 +1,9 @@
+import { ProfilesState } from '../reducers';
+import { createSelector } from 'reselect';
+
+export const getContextPopinState = (state: ProfilesState) =>
+  state.contextPopin;
+export const getContextPopinContent = createSelector(
+  [getContextPopinState],
+  content => content
```
JalilArfaoui, line 8: Shouldn’t that be `({ content }: ContextPopinState) => content` ?

src/app/profiles/App/ContextPopin/withConnect.ts
```diff
@@ -0,0 +1,17 @@
+import { ProfilesState } from '../../store/reducers';
+import { getContextPopinContent } from '../../store/selectors/contextPopin';
+import { setContextPopin } from '../../store/actions/contextPopin';
+import { ContextPopinState } from '../../store/reducers/contextPopin.reducer';
+import { Dispatch } from 'redux';
+import { connect } from 'react-redux';
+
+const mapStateToProps = (state: ProfilesState) => ({
+  contextPopin: getContextPopinContent(state)
```
JalilArfaoui, line 9: I guess what you want is `getContextPopinState` ?

src/app/profiles/store/selectors/contextPopin.ts
gregoirelacoste: you'r right

src/app/profiles/App/ContextPopin/withConnect.ts
gregoirelacoste: hum, I need the contents here so getContextPopinContent

JalilArfaoui (approved):

PR #973, opened 2021-05-13
Feat/fetch and store matching contexts per contributor #953

### LINKED ISSUE #953: Fetch and store matching contexts per contributor
Following our discussion with @JalilArfaoui 
during last staging test we discovered some notices that didn"t triggered in the following cas : 
1. user clicks on see an example of a contributor
2. clicks on "follow"
3. Clicks on "see example"

JalilArfaoui:
Fixes #953 

We don’t want to fetch whole matching contexts every time we subscribe or unsubscribe to a contributor. 

Instead, this implements a fetch on a contributor-basis, a storage with associatedContributors (including relayed matching contexts), and fetch-less unsubscriptions.


JalilArfaoui:
@MaartenLMEM here’s a staging build (Firefox+Chromium) to test :


[bulles-v3.73.2-staging.zip](https://github.com/dis-moi/extension/files/6473458/bulles-v3.73.2-staging.zip)

src/app/background/sagas/refreshMatchingContexts.ts
```diff
@@ -36,29 +35,42 @@ const callAndRetry = createCallAndRetry({
...
+}
+
 export default function* refreshMatchingContextsRootSaga() {
-  yield refreshMatchingContextsSaga();
-  yield debounce(
-    30_000,
-    [SUBSCRIBE, UNSUBSCRIBE, REFRESH_MATCHING_CONTEXTS],
-    refreshMatchingContextsSaga
-  );
+  yield fork(refreshMatchingContextsPeriodically);
```
lutangar, line 74: This is probably enough for our needs, but I probably would have kept an action triggering a refresh..
1. A command asking to refresh/fetch (this way the app know we are `fetching`)
2. Then dispatch another command for each contributor... and wait for all of them to comeback

A bit like the collection system implemented on the profile side.
But I know this is quite an heavy system and might be too much for our needs.

lutangar (approved):

MaartenLMEM:
Hi there, I can use file attached. Can we merge for tomorrow afternoon ?
image

src/app/background/reducers/resources/matchingContexts.reducer.ts
```diff
@@ -1,14 +1,79 @@
...
-export default (state: MatchingContexts = initialState, action: AppAction) => {
+const overAssociatedContributors = (fn: R.Arity1Fn) =>
+  R.over(R.lensProp('associatedContributors'), fn);
+
+export default (
+  state: MatchingContextsState = initialState,
+  action: AppAction
+): MatchingContextsState => {
   switch (action.type) {
     case UPDATE_MATCHING_CONTEXTS:
```
gregoirelacoste, line 45: without going into detail, Webstorm tells `Unreachable 'case' branch `

src/app/background/reducers/resources/matchingContexts.reducer.ts
```diff
@@ -1,14 +1,79 @@
...
+        MatchingContextsState,
+        MatchingContextsState
+      >(
+        R.mapObjIndexed(
+          overAssociatedContributors(
+            R.reject(R.equals(unsubscribedContributorId))
+          )
+        ),
+        R.pickBy(R.compose(R.not, R.isEmpty, R.prop('associatedContributors')))
+      )(state);
```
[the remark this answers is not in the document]
gregoirelacoste: Lot of Ramda, it's harder to follow

gregoirelacoste (approved):
It's seems working great ! 

Tested in local 
Old payload for `Update_Matching_Context` was : `Array(1900)`, one time
New payload for each subscription, ex:
```
payload:
   contributorId: 52
   matchingContexts: (74)
```

src/app/background/reducers/resources/matchingContexts.reducer.ts
JalilArfaoui: I know, but he’s wrong.

src/app/background/sagas/refreshMatchingContexts.ts
JalilArfaoui: I tend more and more to prefer having the control of the entire flow centralized in a saga …

Sure, we may need an action to let the UI know that we’re fetching … We don’t know what to do with this information right now, but the day we do it’s super easy to do : We just `put` a `REFRESHING_MATCHING_CONTEXT` **event** instead of `put`ing a **command** that we wait for in another subsaga …

It gives us even more control, with the ability to `put` intermediate actions, in order to notify about retries, easily parallelize requests (what I didn’t do yet), …

Also, doing this one, the whole fetching MC system is a bit more isolated and independent from the rest of the system

src/app/background/reducers/resources/matchingContexts.reducer.ts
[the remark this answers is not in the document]
JalilArfaoui: I’ll propose some clarification refactoring(s) later, since this has already been merged

PR #976, opened 2021-05-14
fix(onBoardingRestart): memoize onboarding

### LINKED ISSUE #965: Onboarding restarts after I quitted it when I come back on list of profiles
see video : https://drive.google.com/file/d/1wj_NNzEg_sKD61YaJgbdZs_8B4n2IcHr/view

NB : Staging firefox 3.74.0

gregoirelacoste:
closes #965


lutangar (approved):

JalilArfaoui:
I think this fix is a bit fragile … 

Maybe we should think about pushing the onboarding steps in the url to avoid loosing state, even in case of user refresh

PR #1014, opened 2021-05-28
fix(profile): display link

### LINKED ISSUE #1013: On profile : contribution link doesn't appear

gregoirelacoste:
closes #1013

### DIFF
--- a/src/app/profiles/App/Pages/Profiles/Profile/ProfileNoticeListItem.tsx
+++ b/src/app/profiles/App/Pages/Profiles/Profile/ProfileNoticeListItem.tsx
@@ -104,7 +104,7 @@ export const ProfileNoticeListItem = ({
           <Trans i18nKey={'profiles:notice.pined_on'}>
             <NoticeHighlight>
               Message épinglé sur
-              <NoticeURL>{exampleMatchingUrl}</NoticeURL>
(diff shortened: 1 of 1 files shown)


src/app/profiles/App/Pages/Profiles/Profile/ProfileNoticeListItem.tsx
```diff
@@ -104,7 +104,7 @@ export const ProfileNoticeListItem = ({
           <Trans i18nKey={'profiles:notice.pined_on'}>
             <NoticeHighlight>
               Message épinglé sur
-              <NoticeURL>{exampleMatchingUrl}</NoticeURL>
+              <NoticeURL>{{ exampleMatchingUrl }}</NoticeURL>
```
lutangar, line 107: I might miss something here... :thinking: 
From what I see `NoticeURL` is a `span` and `exampleMatchingUrl` is string so one pair of bracket should be enough to displayt this link. 
Aren't these double bracket meaning we're passing an object here?
```
{
    exampleMatchingUrl: 'www.dismoi.io'
}
```

gregoirelacoste: I thought the same thing at first, but i18n needs an object and we are in a Trans component

PR #1015, opened 2021-05-28
feat(externalLinkHandler): handle chrome path and edreams issue

### LINKED ISSUE #1003: edreams : accept cookies closes Dismoi window
https://www.edreams.fr/
Bubble n° (#1079)

gregoirelacoste:
closes #1003


src/app/content/externalClickHandler.ts
```diff
@@ -28,15 +28,27 @@ export const isHtmlElementInteractive = (element?: HTMLElement | null) =>
...
+  );
+
 export default (store: Store) => (e: MouseEvent) => {
   const state = store.getState();
 
   const element = e.target as HTMLElement;
   const parentElement = element.parentElement;
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
+  const chromePath: string[] | null = e?.path && e.path.map(p => p.localName);
```
lutangar, line 44: Could you put this Chrome specific code in the `isHtmlChromeElementInteractive` function instead?
The function name might need a name and argument change....

Afther a bit of digging on this `path` atribute, there is actually a cross-browser solution see:
- https://stackoverflow.com/questions/39245488/event-path-is-undefined-running-in-firefox
- https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath

By using the cross-browser solution you might be able to simplify this to smthing like:
```
isEventPathInteractive(e.composePath);
```
or 
```
isEventInteractive(e);
```

lutangar (changes requested):
> see comment in diff, there are cross-browser solutions

JalilArfaoui (commented):
Same as @lutangar

lutangar:
ping @gregoirelacoste

test/app/content/externalClickHandler.test.ts
```diff
@@ -9,13 +9,23 @@ const document = dom.window.document;
...
+      isEventPathInteractive([
+        document.createElement('span'),
+        document.createElement('button')
+      ])
+    ).to.be.true;
+  });
+  it('can tell if actual Html elements are interactive with onclick attribute', () => {
     const element = document.createElement('div');
     element.setAttribute('onClick', '() => {}');
+    console.log(element);
```
JalilArfaoui, line 23: `console.log`

JalilArfaoui (approved):

src/app/content/externalClickHandler.ts
```diff
@@ -2,41 +2,16 @@ import { MouseEvent } from 'react';
...
   const state = store.getState();
-
-  const element = e.target as HTMLElement;
-  const parentElement = element.parentElement;
-
-  const interactive =
-    isHtmlElementInteractive(element) ||
-    isHtmlElementInteractive(parentElement);
+  const pathElements = e.composedPath && e.composedPath();
+  const interactive = isEventPathInteractive(pathElements);
```
lutangar, line 14: You should cast `pathElements` as `HtmlElement` instead of overriding the `composedPath` function definition. This is one of the few valid use case for casting:
```ts
const interactive = isEventPathInteractive(pathElements as HTMLElement[]);
```
> https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath

src/libs/utils/isEventPathInteractive.ts
```diff
@@ -0,0 +1,24 @@
...
+  'input',
+  'label',
+  'meter',
+  'optgroup',
+  'option',
+  'select',
+  'textarea',
+  '[onclick]'
+];
+const htmlElementInteractive = (element: HTMLElement) =>
```
lutangar, line 19: You should keep previous name `isHtmlElementInteractive`. This was also more explicit since this is a function and it returns a boolean.

PR #993, opened 2021-05-20
Chore/profiles/explaining video

newick:
We'd like to find a way to explain DisMoi directly from the profile page.

First try with a modale opening the explaining video

linked to #922

Issue #922, opened 2021-04-26
V1.3 Profile page improvments to make 1st more intuitive and convert more visitors to users


MaartenLMEM:
1ères idées :

+ C'est quoi Dismoi 
+ CTA installer visible en header ou en colonne de droite

MaartenLMEM:
Design review made with Marco et Pierrick 10 days ago : https://github.com/dis-moi/extension/issues/922
This week Erick is preparing a first design proposal

MaartenLMEM:
one news issue https://github.com/dis-moi/extension/projects/1 I did not see the first time
the rest is ok

PR #1022, opened 2021-05-31
feat(profiles): Remove old install block when not connected

lutangar:
> see https://github.com/dis-moi/extension/pull/993#issuecomment-851507753 and #922 for background


lutangar:
Could you merge this if approved @gregoirelacoste ?

gregoirelacoste (approved):

PR #1023, opened 2021-05-31
chore(folder-structure): Move i18n, hooks, hocs, domain to libs

lutangar:
> This a step further in #1017 refactor, you should probably review #1017 first


gregoirelacoste (commented):
:heart: 
theme and theme.stories is in `src`, why not in libs ?
Could we rename `lmen` to `domain` ?
`Components` is not a `lib` ?
what to do with wies and httpRequests in root folder it's a beat confusing, no ?

lutangar:
> theme and theme.stories is in src, why not in libs ?
> Components is not a lib ?

I don't want to mess with the design system yet, let's merge all @newick refactor before moving these.

> Could we rename lmen to domain ?

Yeah sure will do!

gregoirelacoste (approved):

newick:
i'd like to make `theme` disappear and replace it with a CSS variables files.

PR #1031, opened 2021-06-02
fix(newBlocksFirefox): change test connected in Profile

gregoirelacoste:
related to #1027
The issues was connected === undefined don't match with connected === false in:
``` js
{connected === false && (
          <ExplainingVideoMessageBox contributor={contributor} />
        )}
```
I fixed with `!connected &&`
A question : in connected reducer initial state is connected = undefined, why undefined and not false ?


lutangar (approved):

lutangar:
>  initial state is connected = undefined

It seems on Firefox behavior is sensibly different, but on Chrome it works as initially planned:
Initially we don't know if are connected or not, thus were are in undefined state until there is connected failed or success event.
This avoid flashing of content that may not appear at all once connected. If initially false, the blocks would appears for a sec and then disappear once connected.

PR #1034, opened 2021-06-03
fix(chunk): onboarding chunk path

gregoirelacoste:
chunk name changed for OnBoarding.chunk.js, we mustn't to update path in `functions.php wp_register_script` when we push in prod


src/app/profiles/App/OnBoarding/OnBoarding.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
```
lutangar, line 1: You should probably `git mv` the old index to this file instead.

lutangar (approved):

PR #1050, opened 2021-06-11
feat(profile): famous contributors

### LINKED ISSUE #998: On profile pages, having a popular profiles list instead of similar one
To add some reinsurance, we'd like to display a list of chosen contributors, instead of the similar one.
We'd use the famous names, some real people.

**Target user**
New users. They'd feel better with know names and real faces

The question is: how to easily manage a list? Could we based on SimilarProfiles component?

gregoirelacoste:
closes #998

@MaartenLMEM you can choose any famous contributors you want by editing famousContributors.json in the order you want 
Be careful to write the name of the contributor exactly as in the backoffice
Could you also check the translation ?


lutangar:
The app configuration is done in `env`, we might want to change this but it should be discussed.
See similar feature for popular contributors (this isn't used but should be on install popin) and preselected contributors.
```
POPULAR_CONTRIBUTORS_IDS=[53, 54, 59, 46, 65]
PRESELECTED_CONTRIBUTORS_IDS=[16, 24, 25, 27, 29, 31, 42, 43, 44, 53, 54, 55, 59, 60, 65, 66, 67, 74]
```

gregoirelacoste:
> The app configuration is done in `env`, we might want to change this but it should be discussed.

The goal on this feature is to create something simple to manage for @MaartenLMEM . So using ids are to much complex for @MaartenLMEM beacause it need to go to backoffice and note ids for each contributors. Name is much simplier because it's dispalyed in front, just copy/past.

Env var is a good option, but we also would like something administrable from github

src/app/profiles/store/selectors/index.ts
```diff
@@ -48,6 +49,20 @@ export const getSimilarContributors = createSelector(
     )
 );
 
+export const getFamousContributors = createSelector(
```
JalilArfaoui, line 52:
```
export const getFamousContributors = createSelector(
  getStatefulContributors,
  statefulContributors => famousContributors.map(
    name => statefulContributors.find(contributor => name === contributor.name)
  )
);
```

OR 

```
export const getFamousContributors = createSelector(
  getStatefulContributors,
  statefulContributors => famousContributors.map(
    id => statefulContributors.find(contributor => id === contributor.id)
  )
);
```

if we change to storing ids …

JalilArfaoui:
I tend to agree with @lutangar here, for the sake of consistency … @MaartenLMEM woud it be OK to store the list as an array of IDs that you would change here : https://github.com/dis-moi/extension/blob/master/.env.production ? 

@lutangar actually, this could even be extracted from this physical file to hosted ENV, to enable deploy less editing …

MaartenLMEM:
>  @MaartenLMEM woud it be OK to store the list as an array of IDs that you would change here : [https://github.com/dis-moi/extension/blob/master/.env.production ? ](https://github.com/dis-moi/extension/blob/master/.env.production%E2%80%AF?%E2%80%AF)

All right for me, need already discussed with Grégoire and Johan.

src/app/profiles/store/selectors/index.ts
gregoirelacoste: >  statefulContributors => famousContributors.map(
>     name => statefulContributors.find(contributor => name === contributor.name)
>   )

It's not working like that because `getStatefulContributors` is fired 2 times : 
- first time to get user page
- second time to get all statefull contributors
i don't know why but it same with getSimilarContributors selector

So if we use map, the first time give an array with empty value > it break the component without lot of test to valid. 
If we use foreach and push an init array empty there is no error and no need more test.

lutangar (changes requested):

src/app/profiles/App/famousContributors.ts
```diff
@@ -0,0 +1,7 @@
+import { ContributorId } from '../../../libs/domain/contributor';
+
+const famousContributors: ContributorId[] | undefined =
+  process.env.FAMOUS_CONTRIBUTORS_PROFILE_SIDEBAR &&
+  JSON.parse(process.env.FAMOUS_CONTRIBUTORS_PROFILE_SIDEBAR);
```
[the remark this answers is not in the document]
lutangar: Even better, instead of `ContributorId` use `Contributor['id']`

JalilArfaoui:
@gregoirelacoste where are we on this one ? Do you need any help ?

MaartenLMEM (approved):

gregoirelacoste:
@JalilArfaoui rebased, could you push ?

.env.lmel
```diff
@@ -6,3 +6,4 @@ PROFILES_ASSETS_PATH=https://lememe.fr
 ACCESSIBLE_CONTRIBUTORS_IDS="[25, 35, 40, 43, 61, 118, 130, 140, 153, 161, 165, 166, 167, 170, 172, 173, 174, 175, 176, 177, 178, 179, 181, 182, 183, 60]"
 PRESELECTED_CONTRIBUTORS_IDS="[25, 43, 166, 167, 174, 60]"
 POPULAR_CONTRIBUTORS_IDS="[53, 54, 59, 46, 65]"
+FAMOUS_CONTRIBUTORS_PROFILE_SIDEBAR=[53, 59, 54, 25, 65]
```
JalilArfaoui, line 9: 53, 59, 54, and 65 are not in the accessible contributors in LMEL … @MaartenLMEM what should be the famous contributors in LMEL ?

JalilArfaoui (changes requested):

.env.lmel
MaartenLMEM: Let's say : 166, 174, 160, 153, 167 @JalilArfaoui

Issue #782, opened 2021-02-17
Question: What events are we tracking in the extension?


christpet:
In Matomo, what are these "events" actually tracking?
![image](https://user-images.githubusercontent.com/32476396/108178829-2f767680-7105-11eb-9453-003cbb4000cf.png)

MaartenLMEM:
@christpet It was @lutangar who built up analytics on matomo. I ll know probably better than Jalil.

MaartenLMEM:
I start to answer, tell me if it's what you expected :
- Suscribe = suscribe to an contributor
- Close : extesnion window closed (beacause user clicked on the cross up right the window)
- Click = user clic on extension
etc.
Or do you want more technical informations ?

christpet:
Fantastic thank you @MaartenLMEM that's exactly what we're looking for @zhinu 

That is a LOT of subscribes in one week... is that expected behavior?

MaartenLMEM:
> That is a LOT of subscribes in one week... is that expected behavior?
Yes and no.
Yes : It counts real suscribing events, even when these are automatic
No : Some extensions "loose their mind" about DisMoi : their local storage forget its subscriptions. So we built up a mechanism that suscribe user to  N default contributors if it has zero subscription.

christpet:
Ah wow, that would explain why the number is so big. Really interesting that the extension has amnesia sometimes 😅

Looking at Matomo, it looks like the majority of subscription events are attributable to just 2 users in Ukraine and Australia. Many times per day they log a session where they add 17-22 subscriptions within 2 seconds. There's our automated mechanism.

It's okay for now, we might be able to get better subscription numbers just by filtering out these 2 users. We can look into it for the future.

Right now I'm focused on the usage stats, which are fantastic. Bravo @MaartenLMEM  and @lutangar . I love that we are able to see exactly the user journey people are taking and how they interact, but completely anonymized.

What does it mean when it says "blue" or "grey" next to a content item?

MaartenLMEM:
> Looking at Matomo, it looks like the majority of subscription events are attributable to just 2 users in Ukraine and Australia. Many times per day they log a session where they add 17-22 subscriptions within 2 seconds. There's our automated mechanism.

Look out Chrisitian, those are anonymized users ! They are all differents, and they are in France. Ukraine or Australia is just a way for matomo to anonymize data.

christpet:
Oh wow I didn't realize that even the country is scrambled. Thank you for the heads up!

[...]

PR #1062, opened 2021-06-18
feat(tracking): Add a flag to identify an automatic subscription

### LINKED ISSUE #842: Tell apart subscriptions created automatically from subscriptions done manually
**Describe the bug**
Right now inside the extension, occasionally users will inexplicably lose all their subscriptions. So we have a script running that automatically adds a bunch of subscriptions to any account that has zero subscriptions. 

The problem is that these automatic subscriptions are clogging up Matomo and preventing us from knowing the real stats of when and how users subscribe to contributors.

More information here: https://github.com/dis-moi/extension/issues/782 

**Goal**
Figure out how to get numbers (in Matomo, or even short-term in a spreadsheet) that separate "subscriptions created automatically" from subscriptions created through manual actions by users. So that we can know more about how people are actually using the extension.

@zhinu I know you have started to work on this, but please ping people here if you're getting stuck.

lutangar:
Close #842

### DIFF
--- a/src/app/background/store/sagas/subscriptions.saga.ts
+++ b/src/app/background/store/sagas/subscriptions.saga.ts
@@ -3,7 +3,7 @@ import { SagaIterator } from 'redux-saga';
 import {
   ContributorAction,
   STARTUP,
-  subscribe,
+  autoSubscribe,
   SUBSCRIBE,
   subscribed,
   subscribeFailed,
@@ -79,7 +79,7 @@ export function* autoSubscribeSaga() {
   if (nbSubscriptions === 0) {
     yield all(
       PRESELECTED_CONTRIBUTORS_IDS.map(contributorId =>
-        put(subscribe(contributorId, { manual: false }))
+        put(autoSubscribe(contributorId))
       )
     );
   }
--- a/src/app/background/store/sagas/subscriptions.saga.ts
+++ b/src/app/background/store/sagas/subscriptions.saga.ts
@@ -79,7 +79,7 @@ export function* autoSubscribeSaga() {
   if (nbSubscriptions === 0) {
     yield all(
       PRESELECTED_CONTRIBUTORS_IDS.map(contributorId =>
-        put(subscribe(contributorId))
+        put(subscribe(contributorId, { manual: false }))
       )
     );
   }
--- a/src/libs/store/actions/subscription.ts
+++ b/src/libs/store/actions/subscription.ts
@@ -1,31 +1,37 @@
-import { ContributorId } from 'libs/domain/contributor';
+import { Contributor } from 'libs/domain/contributor';
 import { Level } from 'libs/utils/Logger';
 import { ActionMeta, BaseAction, ErrorAction } from '.';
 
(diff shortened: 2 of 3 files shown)


src/app/background/store/sagas/subscriptions.saga.ts
```diff
@@ -79,7 +79,7 @@ export function* autoSubscribeSaga() {
   if (nbSubscriptions === 0) {
     yield all(
       PRESELECTED_CONTRIBUTORS_IDS.map(contributorId =>
-        put(subscribe(contributorId))
+        put(subscribe(contributorId, { manual: false }))
```
gregoirelacoste, line 82: I think that it would be clearer to create a new action: `autoSubscribe(contributorID)`

src/libs/store/actions/subscription.ts
```diff
@@ -1,31 +1,37 @@
...
 export const SUBSCRIBE = 'SUBSCRIBE';
-export const subscribe = (contributorId: ContributorId, meta?: ActionMeta) => ({
+export const subscribe = (
+  contributorId: Contributor['id'],
+  meta?: ContributorActionMeta
+) => ({
   type: SUBSCRIBE as typeof SUBSCRIBE,
   payload: contributorId,
-  meta
+  meta: { manual: true, ...meta }
```
gregoirelacoste, line 19: it work even with ` meta: { manual: true, ...meta }` ? `put(subscribe(contributorId, { manual: false }))`

gregoirelacoste (changes requested):

src/libs/store/actions/subscription.ts
lutangar: Sorry but I don't understand what you mean :sweat_smile:

gregoirelacoste: i mean you give to the action `{manual: false}`, but it's already by default on `{manual: true}`, so i ask if its still working ?

src/app/background/store/sagas/subscriptions.saga.ts
lutangar: I think it's quite fine to use action `meta` to add precision on the origin of this subscription as first suggested in https://github.com/dis-moi/extension/issues/782#issuecomment-781576687

lutangar: > i mean you give to the action {manual: false}, but it's already by default on {manual: true}, so i ask if its still working ?

The second argument of the subscribe method is an object with a `manual` property set to `false` so it will override  `manual` default value which is `true`.

gregoirelacoste (approved):

src/libs/store/actions/subscription.ts
```diff
@@ -1,31 +1,42 @@
...
 });
 export type SubscribeAction = ReturnType<typeof subscribe>;
 
+export const autoSubscribe = (
```
lutangar, line 23: @gregoirelacoste I added an `autoSubscribe` method per your suggestion

PR #1069, opened 2021-06-24
feat(i18n): sort by language contributors and notices

### LINKED ISSUE #794: Primarily show contributions categorized in the appropriate language (English Version)
** User stories:

- If I am an English-speaking user, I should primarily see contributions and sources that are categorized as English-speaking

gregoirelacoste:
closes #794

### DIFF
--- a/src/libs/i18n/index.ts
+++ b/src/libs/i18n/index.ts
@@ -4,8 +4,10 @@ import { initReactI18next } from 'react-i18next';
 import resources from './resources';
 import format from './format';
 
-export type SupportedLanguage = 'en' | 'fr';
-export const fallbackLng = 'en';
+export type SupportedLanguage = 'fr' | 'en' | 'br';
+export const fallbackLng = 'en' as SupportedLanguage;
+export const defaultLng = 'fr' as SupportedLanguage;
+export const getCurrentLng = () => i18n.language as SupportedLanguage;
 
 export const options = {
   resources,
--- a/test/fakers/generateContributor.ts
+++ b/test/fakers/generateContributor.ts
@@ -8,6 +8,7 @@ import {
   generateNoticeItem,
   generatePinnedNotice
 } from 'test/fakers/generateNotice';
+import { defaultLng } from 'libs/i18n';
 
 interface Options {
   id?: number;
@@ -46,7 +47,8 @@ export const generateContributor = ({
       },
   noticesUrls: ['http://backend.dismoi.io/notices/1'],
   website: website || Faker.internet.url(),
-  categories: [Faker.random.word()]
+  categories: [Faker.random.word()],
+  locale: defaultLng
 });
 
 export const generateStatefulContributor = ({
--- a/src/app/content/store/selectors/index.ts
+++ b/src/app/content/store/selectors/index.ts
@@ -25,6 +25,8 @@ import {
 import { findItemById } from 'libs/utils/findItemById';
 import { StatefulContributor } from 'libs/domain/contributor';
 import { makeGetRouteParam } from 'libs/store/selectors';
+import { getCurrentLng } from 'libs/i18n';
+import sortByLocale from 'libs/utils/sortByLocale';
 import { ContentState } from 'app/content/store/reducers';
 export * from './serviceMessage.selectors';
 export * from './ui.selectors';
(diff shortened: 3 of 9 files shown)


src/libs/i18n/index.ts
```diff
@@ -4,8 +4,10 @@ import { initReactI18next } from 'react-i18next';
 import resources from './resources';
 import format from './format';
 
-export type SupportedLanguage = 'en' | 'fr';
-export const fallbackLng = 'en';
+export type SupportedLanguage = 'fr' | 'en' | 'br';
+export const fallbackLng = 'en' as SupportedLanguage;
+export const defaultLng = 'fr' as SupportedLanguage;
+export const currentLng = () => i18n.language as SupportedLanguage;
```
lutangar, line 10: Why not
```
export const currentLng = i18n.language as SupportedLanguage;
```
like the other defs?

src/libs/utils/arrangeByLanguage.ts
```diff
@@ -0,0 +1,13 @@
+import * as R from 'ramda';
+import { SupportedLanguage } from '../i18n';
+import { Contributor } from '../domain/contributor';
+import { Notice } from '../domain/notice';
+
+function arrangeByLanguage<Items extends Notice | Contributor>(
+  items: Items[],
+  locale: SupportedLanguage
+): Items[] {
+  return R.sort(item => (item.locale === locale ? -1 : 1), items);
```
lutangar, line 10: Ramda don't bring value here:
```
items.sort(item => (item.locale === locale ? -1 : 1))
```

src/libs/utils/arrangeByLanguage.ts
```diff
@@ -0,0 +1,13 @@
+import * as R from 'ramda';
+import { SupportedLanguage } from '../i18n';
+import { Contributor } from '../domain/contributor';
+import { Notice } from '../domain/notice';
+
+function arrangeByLanguage<Items extends Notice | Contributor>(
```
lutangar, line 6: `sortByLanguage` would better describe what it does... or mb `sortByLocale` since we're sorting by locale?

test/fakers/generateContributor.ts
```diff
@@ -8,6 +8,7 @@ import {
   generateNoticeItem,
   generatePinnedNotice
 } from 'test/fakers/generateNotice';
+import { defaultLng } from '../../src/libs/i18n';
```
lutangar, line 11: prefer absolute notation.

src/libs/utils/arrangeByLanguage.ts
```diff
@@ -0,0 +1,13 @@
+import * as R from 'ramda';
+import { SupportedLanguage } from '../i18n';
+import { Contributor } from '../domain/contributor';
```
lutangar, line 3: since you're importing a lot of domain related code... mb your code is also part of the domain :thinking:

src/libs/i18n/index.ts
lutangar: I find it crazy but.... even in i18next the words: locale language (or lng) are used interchangeably, I find it pretty confusing since what's defined as a language in i18next seems to support locale.
Dunno about us, I'd love if we could find a definition on when to use one or the other... @JalilArfaoui

lutangar (changes requested):

src/libs/i18n/index.ts
gregoirelacoste: @lutangar  i18n.language initializes too early and if I do not restart the i18n.language when I use it (with this function) the result may be wrong for the entire app

src/libs/utils/arrangeByLanguage.ts
gregoirelacoste: @lutangar I agree, but it was to keep consistency with the rest, I also prefer to change

src/libs/utils/arrangeByLanguage.ts
gregoirelacoste: hum, it's especially useful but there is no domain logical, no ?

src/libs/i18n/index.ts
lutangar: Alright, let's suppose will fix this issue later, but then do name this as a function instead for example:
```
getCurrentLng
```

gregoirelacoste: your right

lutangar (approved):

PR #1074, opened 2021-07-01
feat(tracking): Add browser, platform and devices info

lutangar:
Retrieve info as PostHog would do on the browser side and update doc.

> Note: utils.ts in tracking libs, is taken from PostHog, I therefore declined all responsability about its content :p

### DIFF
--- a/src/libs/tracking/utils.ts
+++ b/src/libs/tracking/utils.ts
@@ -1,0 +1,163 @@
+// @see https://raw.githubusercontent.com/PostHog/posthog-js/master/src/utils.js
+type BrowserInfo = browser.runtime.BrowserInfo;
+
+type Browser =
+  | 'Opera Mini'
+  | 'Opera'
+  | 'BlackBerry'
+  | 'Internet Explorer Mobile'
+  | 'Samsung Internet'
+  | 'Microsoft Edge'
+  | 'Facebook Mobile'
+  | 'Chrome'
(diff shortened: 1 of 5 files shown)


src/libs/tracking/utils.ts
```diff
@@ -0,0 +1,164 @@
...
+    return '';
+  }
+};
+
+/**
+ * This function detects which browser version is running this script,
+ * parsing major and minor version (e.g., 42.1). User agent strings from:
+ * http://www.useragentstring.com/pages/useragentstring.php
+ */
+const getBrowserVersion = (userAgent: string, vendor: string) => {
```
gregoirelacoste, line 85: you don't want use a module like: https://www.npmjs.com/package/useragent?
It's could be easier to maintain

gregoirelacoste (approved):

src/libs/tracking/utils.ts
lutangar: I wish but... I'm using what's __PostHog__ is using, to be coherent with the values __PostHog__ is using...
And this method is only used as a polyfill for Chrome since Firefox web extension runtime comes with a function doing juste dat already: 
https://developer.mozilla.org/fr/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo

> also about the library we might use only 1% of its features

JalilArfaoui (approved):

Issue #761, opened 2021-01-12
Project structure


lutangar:
Current project structure is described [here] (https://github.com/dis-moi/extension/blob/master/docs/CONTRIBUTING.md#structure)
> btw I'm noticing there's a typo, `app` folder  isn't in the `api` folder of course

### Apps
It was discussed a few times before but here what's I'm suggesting:
The app folder should **only** contains:
- background
- profile
- profiles

...and that's it!

### Store stuff
Everything else should live inside the `src` directory.
That being said, common `redux` (or `store` I don't really care) should live inside a dedicated directory.
For example assuming we are moving `src/app/store` to  `src/store` let's put inside the following folder:
- app/actions
- app/sagas

@felix-lambert @JalilArfaoui thoughts?

> Please also ping any other persons with a frontend sensibility I might have forgot...

JalilArfaoui:
`src/app` could event be named `src/apps` :-)

We should have 
- src/store/actions
- src/store/sagas
- src/store/…
- src/apps/content/store/… for specific stuff …

lutangar:
> https://nx.dev/latest/react/getting-started/nx-setup#folder-structure

Let's see if we can move one step closer to `nx` folder structure.

gregoirelacoste:
@lutangar  to summarize our talk:
- first step: we reorganize folders structure with @JalilArfaoui 
- second step: act if we use or not nx ?
- third step: implementation ?

JalilArfaoui:
Yep !

MaartenLMEM:
Please for next steps here, try to evaluate _en gros_ how much time current team will gain with new work here.

JalilArfaoui:
@MaartenLMEM to me, given that : 
 - https://github.com/dis-moi/extension/pull/949 is not ready yet, with a history conservation issue in particular, there is some hours to spend on it I guess
 - Current team is stable and already paid the cost of a newcomer being confused about project structure
 - Website V2 is less ambitious than projected 

I’d argue that this - a global one step structure refactoring - is not our priority right now … 

Instead, improving the folder structure with the time, small steps in each PR, can lead us to a posisiton where this global refactoring is easy, if not already done, and the eventual move to nx cheaper. 

IMHO, I’d keep this issue opened for reference, in `Backlog`, and close #949 (that would need rebasing anyway)  for now

gregoirelacoste:
[...]

PR #1084, opened 2021-07-07
chore: update ESLint setup and remove git hooks

JalilArfaoui:
I’ve tried to reduce git hooks time … it seems Prettier take a long time, and also ESLint (> 10 sec), even on a small subset of files (> 5 sec).
I think it’s due to TypeScript that need to transpile everything, but also to the number of active rules we have …
I’ve updated all the associated modules, tried different setups, searched GH issues …

The more I think about it, the more I’d advocate to remove from git hooks at least all checks that IDE can give live feedback, and that are also verified in CI : Prettier, ESLint, typecheck … We already have optional immediate feedback, and safety net …

Actually, I would even go a step further and remove also unit tests (so all git hooks) because I’m not fan of system deciding for me to run tests without event knowing what I changed and if there is any risk to break those tests … and we always have CI to keep things right …

Since it seems half of us commit and push with `--no-verify` (because of the time it takes), the purpose is defeated anyway …

Of course, we do need a ruthless CI …

Tell me what you think, especially @lutangar


lutangar (commented):
While I agree these hooks are now a pain, it's only because they are too long to execute...
I would rather keep, at least, some of them and try to reduce their execution time instead of only getting rid of them... 

I firmly believe (better) project scoping  will help speeding processes, mainly because we won't have to test the whole app if we're working, for instance, on the profiles app only.

I wondered if maybe we could only keep running hooks on push but that would potentially means having one commit failing the tests or cscheck and I don't want that to happens:  __every commit should run__ (or maybe if it's a too heavy constraint, we -- maintainers -- should `squash` failing commits before merging @newick  PRs for example)

I have second thought about prettier and the linter... are the hooks still running only on the staged files since they have path as arguments? 

Anyway, I originally didn't want to spend too much time on tooling while #761 is not fully implemented (nx).

I'm not against merging this PR for the time being, but:
1. maybe we should better document command to clean our code automatically since 
human  will need to think about running it now.
2. strategy may be subject to changes in the near future

lutangar:
> mainly because we won't have to test the whole app if we're working, for instance, on the profiles app only.

See `affected` in `nx` for example https://nx.dev/latest/react/getting-started/intro#features

PR #1086, opened 2021-07-08
Feat/service message

### LINKED ISSUE #1066: News service message
Ce qui s'est passé le mois dernier : N nouveaux contributeurs, La contribution la plus lue, les nouvelles features, etc.,

gregoirelacoste:
closes #1066

### DIFF
--- a/src/app/background/store/sagas/news.saga.ts
+++ b/src/app/background/store/sagas/news.saga.ts
@@ -1,0 +1,20 @@
+import { put } from 'redux-saga/effects';
+import { createErrorAction } from 'libs/store/actions/helpers';
+import { Level } from 'libs/utils/Logger';
+import { showNews } from '../../../../libs/store/actions/news.actions';
+import Tab from '../../../../libs/domain/tab';
+
+export function* showNewsSaga(tab: Tab) {
+  try {
+    yield put(
+      showNews(
+        {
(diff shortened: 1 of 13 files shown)


gregoirelacoste:
@newick could you give me a hand for the animation of this component src/components/organisms/Notification/NotificationNews/index.tsx 
I would like to create a horizontal scroll animation to display long text.
I have issue to scroll 100% of the width of the box, I don't understand why, do you have an idea ?

newick:
Ok I made few tests.
with the current code we can display the whole text BUT:
- time must vary with text length
- we should find a way to loop it better to improve readibility

Can JS do something for us to have a text-long agnostic component?

JalilArfaoui:
@gregoirelacoste is this still WIP ?

gregoirelacoste:
> Ok I made few tests.
> with the current code we can display the whole text BUT:
> 
>     * time must vary with text length
> 
>     * we should find a way to loop it better to improve readibility
> 
> 
> Can JS do something for us to have a text-long agnostic component?



> Can JS do something for us to have a text-long agnostic component?

@newick just a little bit math `const velocity = (news?.length || 0) / 10;`

gregoirelacoste:
> @gregoirelacoste is this still WIP ?

not anymore

PR #1087, opened 2021-07-13
fix: fallback lng not recognized on init

gregoirelacoste:


lutangar (approved):
Grrr... I thought I fixed the circular dependency but apparently not -_-
Let's merge this quickfix n talk about it later...

PR #1096, opened 2021-07-21
fix: list contributors ordering

### LINKED ISSUE #1095: Oddly, profiles on proding are still displayed in increasing order
Firefox 89.0.2 - Proding : https://proding-profiles.dismoi.io/eclaireurs

! Bug Not met on chrome !

Observed with extension and without, incognito mode and normal browsing mode

image

### LINKED ISSUE #1097: After english contributors, French should be diplayed in decreasing order
see https://github.com/dis-moi/extension/issues/794#issuecomment-884172031

Probably same issue as https://github.com/dis-moi/extension/issues/1095

gregoirelacoste:
closes #1095
closes #1097

### DIFF
--- a/src/libs/utils/sortByLocale.ts
+++ b/src/libs/utils/sortByLocale.ts
@@ -1,4 +1,3 @@
-import * as R from 'ramda';
 import { SupportedLanguage } from '../i18n';
 import { Contributor } from '../domain/contributor';
 import { Notice } from '../domain/notice';
@@ -7,7 +6,9 @@ function sortByLocale<Items extends Notice | Contributor>(
   items: Items[],
   locale: SupportedLanguage
 ): Items[] {
-  return R.sort(item => (item.locale === locale ? -1 : 1), items);
+  const localItems = items.filter(i => i.locale === locale);
+  const extItems = items.filter(i => i.locale !== locale);
+  return [...localItems, ...extItems];
 }
 
 export default sortByLocale;
--- a/src/libs/domain/contributor.ts
+++ b/src/libs/domain/contributor.ts
@@ -103,7 +103,7 @@ export const sortContributorsByContributions: SortSuggestedContributors = R.sort
 export const sortContributorsBySubscribers = (
   contributors: Contributor[]
 ): Contributor[] =>
-  R.sort((a, b) => a.ratings?.subscribes - b.ratings?.subscribes, contributors);
+  R.sort((a, b) => b.ratings?.subscribes - a.ratings?.subscribes, contributors);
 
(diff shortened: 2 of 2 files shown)


src/libs/utils/sortByLocale.ts
```diff
@@ -7,7 +6,9 @@ function sortByLocale<Items extends Notice | Contributor>(
   items: Items[],
   locale: SupportedLanguage
 ): Items[] {
-  return R.sort(item => (item.locale === locale ? -1 : 1), items);
+  const localItems = items.filter(i => i.locale === locale);
+  const extItems = items.filter(i => i.locale !== locale);
+  return [...localItems, ...extItems];
```
lutangar, line 11: Prefer a sort function, something like this for example :
```ts
items.map(i => i.locale === locale ? -1 : 1)
```

lutangar: Hu? It was previously the case? Wonder why ramda was used tho...
Previous solution synthax looks right, but dunno ramda well enough tho'

gregoirelacoste: 2 issues with sort : 
- it change the original array, so that why I used R.sort which do not change the original array
- .`sort` (vanilla and Ramda) give different result in chrome and firefox, asc in firefox, desc in chrome (probably because we sort first time by subscribers and then by language)

Syntax : I prefer this syntax, easier to understand and avoid to use to much sort (by subscriber + locale et more later...)

lutangar: You're right about the mutation, but you could use something like `[...items]` to return a new value.

> Syntax : I prefer this syntax, easier to understand and avoid to use to much sort

This is completely subjective. Using **sort** functions (either Ramda or vanilla) to **sort** items is supposed to be the more natural way. Compare functions may then be composed (using Ramda or other functional compose implementation) to work together.

> asc in firefox, desc in chrome

It's not supposed to happens. However it's true some elements with equals position might change a few position on successive runs even on the same browser (not necessarily because of Chrome or Firefox).
> see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

gregoirelacoste: @lutangar 
> You're right about the mutation, but you could use something like `[...items]` to return a new value.
agree

> It's not supposed to happens
I saw something similar https://forum.freecodecamp.org/t/the-sort-method-behaves-different-on-different-browsers/237221/3

but despite everything, I tell myself that this function should not `sort` Contributors but `split` the contributors (by local, my local in an array and others in an other array) that's why I really prefer this syntax

PR #1100, opened 2021-07-25
chore: rename occurrences of bulles.fr to equivalent dismoi.io domain

lutangar:
> see https://github.com/dis-moi/backend/issues/417
> and related PR https://github.com/dis-moi/backend/pull/426

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,6 +1,6 @@
 BABEL_ENV=production
 
-BACKEND_ORIGIN=https://notices.bulles.fr/api/v3/
+BACKEND_ORIGIN=https://api.dismoi.io/api/v3/
 
 # Refresh intervals in minutes
 REFRESH_MC_INTERVAL=30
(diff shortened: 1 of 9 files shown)


.env.production
```diff
@@ -1,6 +1,6 @@
 BABEL_ENV=production
 
-BACKEND_ORIGIN=https://notices.bulles.fr/api/v3/
+BACKEND_ORIGIN=https://api.dismoi.io/api/v3/
```
JalilArfaoui, line 3: Do we still have `/api` prefix for V3 ?

JalilArfaoui (approved):

.env.production
MaartenLMEM: Jalil you approved these changes, but I read your comment. 
is it really ok for you ?

JalilArfaoui: I wasn’t sure because I didn’t verify myself … that’s why I asked the question, to be sure @lutangar spots it … If he says «yes we still have the prefix on V3» then he’ll know he can merge, if not, when he will update the PR :-)

My comment was more a highlight if you prefer

lutangar: > Do we still have /api prefix for V3 ?

Yes prefix is still present, I'd love to see an alias without it but fail to find an easy way to do so on the backend.
Might be an easy pick for you tho'

PR #1124, opened 2021-08-25
chore(safari): setup safari build and first automation

lutangar:
This is a first draft, I still need to figure out the best way to ssh to this machine to trigger macOS and iOS build from our current/future CI.

The `xcrun safari-web-extension-converter [...]` command override every files in target directory including the `.git` folder...
It means I cant first clone the repo and the run the file generation, so I must figure out a git strategy that would preserve other files added to the [safari repository](
https://github.com/dis-moi/extension-safari) (for example a README or a Github workflow, etc...)

### DIFF
--- a/bin/update-safari-repo.sh
+++ b/bin/update-safari-repo.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+cd DisMoi
+git init
+git remote add origin git@github.com:dis-moi/extension-safari.git
+git fetch
+git reset origin/main
+git checkout -t origin/main
+git add .
+git config --global user.name "bullito"
+git config --global user.email "bully@dismoi.io"
+git commit -m "update build"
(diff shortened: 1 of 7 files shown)


bin/update-safari-repo.sh
```diff
@@ -0,0 +1,12 @@
...
+git init
+git remote add origin git@github.com:dis-moi/extension-safari.git
+git fetch
+git reset origin/main
+git checkout -t origin/main
+git add .
+git config --global user.name "bullito"
+git config --global user.email "bully@dismoi.io"
+git commit -m "update build"
+git push origin HEAD
```
JalilArfaoui, line 12: Why not just `git push origin main` ?

JalilArfaoui (approved):
Nice start

PR #1129, opened 2021-08-30
fix: bad xpath expression breaks matching for everyone

JalilArfaoui:
fixes #1127 

Problem was :
 - We filter in all matching contexts the ones which match current URL
 - We filter again the remaining list to check if an xpath expression applies
 - One notice has a bad xpath -> it breaks the whole matching for all notices, even those without XPath !
 
This commit  considers an xpath as non matching if the evaluation fails, and prints a warning to the website console.

PR #1130, opened 2021-08-30
Perf/filter matching lists at once

JalilArfaoui:
To review once #1129 is merges, as this one is based on it … (or review only commit 09d67e5ed4c1071d2a2f19fcf12d291a2eae9829)

This is a little perf refacto that aim to filter matching contexts once instead of first filter for urlRegex and then filtering again for XPath …

### DIFF
--- a/src/libs/domain/matchingContext.ts
+++ b/src/libs/domain/matchingContext.ts
@@ -37,14 +37,6 @@ export const urlMatchesContext = (
   }
 };
 
-export const filterContextsMatchingUrl = (
-  url: string,
-  matchingContexts: MatchingContext[]
-) =>
-  matchingContexts.filter((context: MatchingContext) =>
-    urlMatchesContext(url, context)
-  );
-
 export const doesTabContentMatchExpression = async (
   tab: Tab,
   xpath: string,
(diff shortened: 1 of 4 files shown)


src/libs/domain/matchingContext.ts
```diff
@@ -57,18 +49,28 @@ export const doesTabContentMatchExpression = async (
...
+  tab: Tab
+): Promise<MatchingContext[]> => {
   const responses = await Promise.all(
-    matchingContexts.map(({ xpath, id }: MatchingContext) => {
-      return (
-        typeof xpath === 'undefined' ||
-        doesTabContentMatchExpression(tab, xpath, id)
-      );
+    matchingContexts.map((matchingContext: MatchingContext):
+      | Promise<MatchingContext | false>
```
lutangar, line 58: Could the return of this map be simplified to something simpler like `Promise<MatchingContext | null>` ?

lutangar (approved):

src/libs/domain/matchingContext.ts
JalilArfaoui: Sorry I did not see your comment @lutangar

Done !

PR #1136, opened 2021-09-07
feat(contribution): new no notice screen

### LINKED ISSUE #1126: As a user I can directly contribute from the NoNotice screen
Instead of choosing between contribute and submit a question, NoNoticeScreen would be like:
- a textearea and a submit button
- next screen: user name and user email and a preview button
- same preview screen as in http://storybook.lmem.net/?path=/story/extension-contribute-02-preview--normal
edit button would send back on the first screen (I don't recall doing something else for the actual workflow)
- same submitted screen as in http://storybook.lmem.net/?path=/story/extension-contribute-submitted--03-submitted

gregoirelacoste:
closes #1126
For you @newick 
@lutangar I have a typescript issue, i have to push --no-verify, could check ?

### DIFF
--- a/src/app/background/index.ts
+++ b/src/app/background/index.ts
@@ -30,12 +30,18 @@ if (NODE_ENV !== 'production') {
 console.info(`BACKEND_ORIGIN "${BACKEND_ORIGIN}"`);
 
 onInstalled.then(installedDetails =>
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // @ts-ignore
   store.dispatch(installed(installedDetails))
 );
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
 onStartup.then(() => store.dispatch(startup()));
-
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
 i18n.init(options).then(() => store.dispatch(i18nReady()));
-
+// eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+// @ts-ignore
 const handleConnect = (port: Port) => store.dispatch(connect(port));
 
 browser.runtime.onConnectExternal.addListener(handleConnect);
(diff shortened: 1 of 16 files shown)


lutangar:
I'm on it.

src/app/content/App/Contribute/ContributeScreen/SubmitContributionForm/Step1.tsx
```diff
@@ -0,0 +1,37 @@
+import { Field } from 'redux-form';
+import React, { ChangeEvent, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { TextareaField } from '../../../../../../components/organisms/Fields';
```
lutangar, line 4: Check for unnecessary relative imports.

lutangar:
Looks fine on my end and on the CI, probably because your local dependencies are different, since you're working on the website branch at the same time.

About `redux-form`, for multiple steps form here's the idiomatic way :
https://redux-form.com/6.0.0-alpha.6/examples/wizard/

I thought it was your isssue (wrap each component  inside a `withReduxForm` decorator) but apparently it's not so don't pay too much attention to this.

lutangar (approved):

gregoirelacoste:
@lutangar thx for your feedback

src/app/background/index.ts
```diff
@@ -30,12 +30,18 @@ if (NODE_ENV !== 'production') {
 console.info(`BACKEND_ORIGIN "${BACKEND_ORIGIN}"`);
 
 onInstalled.then(installedDetails =>
+  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
lutangar, line 33: You probably had to do this because you didn't have the right dependencies for this branch.
You should run `yarn install`  and these errors would probably go away.

gregoirelacoste: I did, remove yarn.lock and node_modules then yarn install  but still the same issues...

lutangar: You shouldn't remove your `yarn.lock`!

gregoirelacoste: to renew dependecies it's important no ?

newick (approved):

gregoirelacoste:
@lutangar i have some ts issues from bad node modules dependencies even if i removed the bad yarn.lock and replace by the one from master, could you help me please ?

PR #1143, opened 2021-09-23
fix(useContributorFilter): reset contributor on change tab

### LINKED ISSUE #1141: Proding 3.88.0 : could not follow neither unfollow a contributor on his contributor page during many seconds - it was only possible on page listing profiles
Probably linked to https://github.com/dis-moi/extension/issues/1132 
I close and we will see if it back after #1132 is solved

gregoirelacoste:


src/app/profiles/hooks/useContributorsFilters.ts
```diff
@@ -24,7 +24,13 @@ function useContributorsFilters(
     );
 
   useEffect(() => {
-    if (selectedCategory && selectedCategory === ALL) {
+    console.log(
```
lutangar, line 27: We shouldn't log anything in the content script.
We may log things in the background using the appropriate `logger` service.

lutangar (changes requested):

JalilArfaoui (approved):
Seems fine... Apart from the forgotten `console.logs`

PR #1145, opened 2021-09-27
fix(profilesList): changes list item width

### LINKED ISSUE #1144: Make long local profile names readable on the page listing profiles - Le Même en Nord-Pas-de-Calais etc.
See https://proding-profiles.dismoi.io/eclaireurs profiles beginning with `Le même en ...`
Challenge here is that if we resize one part we ll have to rezise others.

newick:
closes #1144

### DIFF
--- a/src/components/atoms/Contributor/ContributorInfos.ts
+++ b/src/components/atoms/Contributor/ContributorInfos.ts
@@ -1,5 +1,4 @@
 import styled from 'styled-components';
-import UserName from '../UserName/UserName';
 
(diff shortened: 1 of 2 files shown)


JalilArfaoui:
I don’t reproduce locally even after a `rm -rf node_modules && yarn install --frozen-lockfile` but I always reproduce on CI … 

I’m trying to make sure we use the same version of Node

JalilArfaoui:
Fixed !

 I merge it

PR #1165, opened 2021-10-30
Feat/create lmel theme

### LINKED ISSUE #1159: Create LMEL Styled-components theme, with inheritence from default DisMoi theme

JalilArfaoui:
Use the FACET env var (`dismoi` or `lmel`) to apply the correct theme (see /libs/facets/theme) in extension and profiles, and add theme knob in Storybook

`lmelTheme` inherits from `dismoiTheme` … you only have to define things that are different @newick 

Closes #1159


src/app/profiles/App/index.tsx
```diff
@@ -2,25 +2,27 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import { ConnectedRouter } from 'connected-react-router';
 import { ThemeProvider } from 'styled-components';
-
-import theme from 'app/theme';
+import { useTheme } from 'libs/facets/useTheme.hook';
```
lutangar, line 5: :+1:

src/app/profiles/App/index.tsx
```diff
@@ -2,25 +2,27 @@ import React from 'react';
...
-          <PageContainer>
-            <Pages />
-          </PageContainer>
-        </Background>
-      </ConnectedRouter>
-    </ThemeProvider>
-  </Provider>
-);
+const App = () => {
+  const theme = useTheme();
```
lutangar, line 12: :+1:

src/components/atoms/Title/Title.ts
```diff
@@ -1,5 +1,5 @@
 import styled from 'styled-components';
-import { Theme } from '../../../app/theme';
+import { Theme } from '../../../libs/facets/theme';
```
lutangar, line 2: I'm not convinced `Theme` definition should live nor in facets nor in libs.... could it be kept in components?
Could our components decently avoid depending on our libs? Might not be practical tho'

lutangar (approved):
Nice one :+1:

src/components/atoms/Title/Title.ts
lutangar: Also, prefer an absolute import over a relative one.

JalilArfaoui: If we go full lib style, we could have 
- /libs/facet <- /components depends on this lib
- /libs/dismoiFacet <- depends on /libs/facet
- /libs/lmelFacet <- depends on /libs/facet

but for a few files and 2 themes, I thougth I’ll keep it simple for now …

JalilArfaoui: relative import fixed

Issue #1155, opened 2021-10-21
Diverge i18n resources with DisMoi and LMEL sections


JalilArfaoui:

J-oh-n:
> @J-oh-n @newick Do you have an idea which texts are going to be different between DisMoi and LMEL facets in https://github.com/dis-moi/extension/blob/master/src/libs/i18n/resources/fr/extension.json ?
> 
> I’m not sure we’ll have that much differences …

=> we will see it during testing, you can move on

prk-dismoi:
Hey @JalilArfaoui can I move this one to "Ready for prod"?

JalilArfaoui:
Yes

JalilArfaoui:
Yes

PR #1166, opened 2021-10-30
feat: Adapt manifest.json to LMEL facet

### LINKED ISSUE #1162: Adapt manifest.json to LMEL facet

JalilArfaoui:
Closes #1162

### DIFF
--- a/src/components/molecules/Popin/PopinBottomBar.tsx
+++ b/src/components/molecules/Popin/PopinBottomBar.tsx
@@ -9,6 +9,7 @@ import LogoLesInrocks from 'assets/img/contributors/logo-les-inrocks.png';
 import LogoMicode from 'assets/img/contributors/logo-micode.png';
 import LogoQueChoisir from 'assets/img/contributors/logo-que-choisir.png';
 import { WEBSITE_DOMAIN } from 'libs/domain';
+import { useFacetName } from 'libs/facets/useFacetName.hook';
 
 const Link = styled(ExternalLink)`
   color: ${props => props.theme.text};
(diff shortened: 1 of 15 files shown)


src/components/molecules/Popin/PopinBottomBar.tsx
```diff
@@ -9,6 +9,7 @@ import LogoLesInrocks from 'assets/img/contributors/logo-les-inrocks.png';
 import LogoMicode from 'assets/img/contributors/logo-micode.png';
 import LogoQueChoisir from 'assets/img/contributors/logo-que-choisir.png';
 import { WEBSITE_DOMAIN } from 'libs/domain';
+import { useFacetName } from '../../../libs/facets/useFacetName.hook';
```
lutangar, line 12: Prefer absolute path.

src/libs/facets/dismoi.facet.ts
```diff
@@ -0,0 +1 @@
+export const facetName = 'DisMoi';
```
lutangar, line 1: I think this would be best kept as an env variable. 
As I understand you don't want to have too many env variables. However, one possible way to overcome this potential issue, could to create an env variable with a complex structure.
We already have some array for example. So might have a simple js object or a json one if you prefer.
Or would it make sense to create another config file? For example let's look at a silly example:
```
# [...]
THEME_PATH=lmem.theme.json
```

The file idea might be too much :D but think about the idea of a more complex data structure kept in env....

lutangar: I also noticed `WEBSITE_DOMAIN` should be configurable as well.

lutangar: Yet another point... some translation keys are too specific and should be kept "facet" agnostic  like for example :
`profiles:action.know_more_dismoi`

lutangar (changes requested):

src/libs/facets/dismoi.facet.ts
JalilArfaoui: Since our app code will be aware of the facets (for example we will have things like :
```tsx
return facet === 'lmel' ? <This /> : <That />
```
), I think it defeats the point of having it all as env vars … Our app will be aware of facet particularities anyway … don’t you think ?

JalilArfaoui: About the translations keys, it’s being worked on here : #1155

PR #1168, opened 2021-11-08
Feat/limit lmel to local contributors

### LINKED ISSUE #1149: Default subscriptions on LMEL app and extension
Default value for LMEL will be different from DisMoi. 
Yet feature is the same, managed in a file that we can change here throught a Github file.

### LINKED ISSUE #1151: Get only Local contributors on LMEL extensions (Chrome & Firefox)
Currently Dismoi browser extension gets 
- all contributors with minimum 1 bubble
- all public bubbles 

LMEL browser extension just needs 
- local contributors 
- bubbles they made
- bubbles they relayed

**Technical implementation notes**
Any ideas you have about how this could be implemented.

**Additional context**
Will follow https://github.com/dis-moi/backend/issues/458

JalilArfaoui:
Closes #1149 
Closes #1151


lutangar (changes requested):
In my humble opinion LMEL is yet another environment...
So I think it doesn't make much sense to duplicate the env variables, we should rather create yet another env file.

lutangar:
*...thinking out loud here but...* 
...forking this repository to put some specific LMEL code *might* make more sense than adding some of those specific here... :thinking:

JalilArfaoui:
> In my humble opinion LMEL is yet another environment... So I think it doesn't make much sense to duplicate the env variables, we should rather create yet another env file.

Even if it’s common practice, I’m not sure about having all those .env files anyway (https://github.com/motdotla/dotenv#faq) … But that is true that overriding values feels better than creating multiples ones … So I created à `env.lmel` that overrides others `.env` files … check it out

JalilArfaoui:
> _...thinking out loud here but..._ ...forking this repository to put some specific LMEL code _might_ make more sense than adding some of those specific here... thinking

A fork that we would have to keep in sync in the future … would that still be a fork ?

loadEnv.js
```diff
@@ -4,8 +4,13 @@ const dotenv = require('dotenv');
 
 function loadEnv({ path, encoding, debug }) {
   const NODE_ENV = process.env.NODE_ENV || 'development';
+  const FACET = process.env.FACET || 'dismoi';
```
lutangar, line 7: :+1:

src/libs/store/sagas/refreshContributors.saga.ts
```diff
@@ -5,6 +5,21 @@ import {
...
+import { asArray } from '../../utils/env';
+
+const ACCESSIBLE_CONTRIBUTORS_IDS = asArray<ContributorId>(
+  process.env.ACCESSIBLE_CONTRIBUTORS_IDS
+);
+
+const filterAccessibleContributors = (
+  contributors: Contributor[]
+): Contributor[] =>
+  ACCESSIBLE_CONTRIBUTORS_IDS &&
```
lutangar, line 18: Please do use the `asArray` env helper method. (or enhance it if the current method isn't sufficient)
> https://github.com/dis-moi/extension/blob/master/src/libs/utils/env.ts

JalilArfaoui: You’re right … done

PR #1177, opened 2021-11-10
fix: date format on extension closes #1105

### LINKED ISSUE #1105: Repair date format in notice detailed view
L'affichage de la date a sauté à nouveau pour une forme peu lisible.  Observé sur Chrome prod et proding Version 3.86.7

Nous avions rencontré le même problème en mai : https://github.com/dis-moi/extension/issues/927

image

gregoirelacoste:
closes #1105

### DIFF
--- a/src/libs/i18n/format.ts
+++ b/src/libs/i18n/format.ts
@@ -34,7 +34,7 @@ const getFollowersFormat: FormatI18nValueFunction = (value: number) =>
   value > 999 ? (value / 1000).toFixed(1) + 'K' : value;
 
 const getFormat: FormatI18nValueFunction = (value, format, lng) => {
-  if (value instanceof Date) return getDateFormat(value, format, lng);
+  if (parseISO(value) instanceof Date) return getDateFormat(value, format, lng);
   if (format === 'followers') return getFollowersFormat(value);
   return value;
 };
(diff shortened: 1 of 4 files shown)


src/libs/i18n/format.ts
```diff
@@ -34,7 +34,7 @@ const getFollowersFormat: FormatI18nValueFunction = (value: number) =>
   value > 999 ? (value / 1000).toFixed(1) + 'K' : value;
 
 const getFormat: FormatI18nValueFunction = (value, format, lng) => {
-  if (value instanceof Date) return getDateFormat(value, format, lng);
+  if (parseISO(value) instanceof Date) return getDateFormat(value, format, lng);
```
JalilArfaoui, line 37: I’m not sure I agree with the change … 

It seems that the `value` we pass to `getFormat` is always a `Date` in components (even though we do pass strings in unit tests 🤔  …

Example here : https://github.com/dis-moi/extension/blob/d9d838ca1fdb503287c7e0e5a7151152d0625ed3/src/components/organisms/NoticeDetails/NoticeDetails.tsx#L249

Why would we `parseISO` ? 

I suggest:
 - typing `value` parameter
 - change unit tests to pass `Date`s instead of `string`s to be like the app 
 - adding a failing test case for the issue we have

Let’t do it right since it’s not our first date format issue …

JalilArfaoui (changes requested):

PR #1194, opened 2021-12-08
Work on LMEL website (update ui and text)

antoinewolff:
@JalilArfaoui  : I let you check.


src/components/website/molecules/Header/Header.tsx
```diff
@@ -112,7 +120,7 @@ const Header = styled(
         </header>
         <Modal open={modalOpen} setOpen={setModalOpen}>
           <NavMobile>
-            <ListLinks links={links} />
+            <ListLinksMobile links={links} />
```
JalilArfaoui, line 123: Always mobile version ?

antoinewolff: I need to use NavMobileItem instead of NavItem.

PR #1201, opened 2021-12-22
Fix/gatsby build missing style

JalilArfaoui:


lutangar:
@JalilArfaoui don't put us in reviewers if you don't want any reviews. But if you do, left us with more than 1 minute to do so :sweat_smile:

JalilArfaoui:
@lutangar It was more like "hey look what I did... and tell me if you have any comment"... If I don't tag you may not be aware of my change... Sorry if it disturbed you!

PR #1215, opened 2022-01-07
update text for add_follow_box

antoinewolff:


JalilArfaoui (commented):
I’d like to understand what this PR changes ? It seems to me that it’s the same result ?

antoinewolff:
> I’d like to understand what this PR changes ? It seems to me that it’s the same result ?

Because we don't want to display contributor name on LMEL facet, so we can leave it empty.

JalilArfaoui (approved):

PR #1244, opened 2022-03-15
fix(profiles): getPathToContributor does not include path prefix

### LINKED ISSUE #1242: Profile pages on Dismoi.io answer with "Page introuvable"

JalilArfaoui:
Fixes #1242

### DIFF
--- a/src/app/profiles/App/pathToContributor.ts
+++ b/src/app/profiles/App/pathToContributor.ts
@@ -1,12 +1,16 @@
 import slugify from 'slugify';
 import i18n from 'i18next';
+import join from 'join-url';
 import { Contributor } from 'libs/domain/contributor';
 import { getPathPrefix } from '../../../libs/facets/getPathPrefix';
 
 const pathToContributor = (contributor: Contributor) => {
-  return `${getPathPrefix(i18n)}${i18n.t('path.profiles.contributors')}/${
-    contributor.id
-  }/${slugify(contributor.name)}`;
+  return join.pathname(
+    getPathPrefix(i18n),
+    i18n.t('path.profiles.contributors'),
+    contributor.id,
+    slugify(contributor.name)
+  );
 };
 
 export default pathToContributor;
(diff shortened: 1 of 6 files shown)


src/libs/facets/getPathPrefix.ts
```diff
@@ -0,0 +1,5 @@
+import { i18n } from 'i18next';
+import { getFacet } from './getFacet';
+
+export const getPathPrefix = (i18n: i18n): string =>
+  getFacet() === 'lmel' ? '' : `/${i18n.language}`;
```
lutangar, line 5: It seems it was this way before BUT feels akward to  have a `/` in one case and none in the other...

src/app/profiles/App/pathToContributor.ts
```diff
@@ -1,11 +1,12 @@
 import slugify from 'slugify';
 import i18n from 'i18next';
 import { Contributor } from 'libs/domain/contributor';
+import { getPathPrefix } from '../../../libs/facets/getPathPrefix';
 
 const pathToContributor = (contributor: Contributor) => {
-  return `${i18n.t('path.profiles.contributors')}/${contributor.id}/${slugify(
-    contributor.name
-  )}`;
+  return `${getPathPrefix(i18n)}${i18n.t('path.profiles.contributors')}/${
```
lutangar, line 7: For complex path joining it would be way cooler to use smthing like node's `path.join`.
Here we don't which part come with a `/` or not...
Maybe smthing could be crafted with the [URL API](https://developer.mozilla.org/fr/docs/Web/API/URL) instead?

src/libs/facets/getPathPrefix.ts
JalilArfaoui: It's because we add a whole segment in one case, and not in the other … `/fr/something` versus `/something` …

src/app/profiles/App/pathToContributor.ts
JalilArfaoui: We can't use `path` since we’re browser side, and `URL` constructor doesn’t seem to have a builin parts joining feature … 

I’m lookinq at https://www.npmjs.com/package/join-path …

JalilArfaoui: Nope, it also needs `path` … I’m looking at https://www.npmjs.com/package/join-url

lutangar: Love it!

